PR #29, opened 2025-06-09
feat(propagation): not recording propagation

iunanua:
# What does this PR do?

- When injecting, if there is no propagation data in the registry, like for the case of a not recording span, try to get the propagation data from `tracestate`.
- When extracting always, even if a `tracestate` is not received, create a `tracestate` with sampling priority, dm and propagation tags.

# Motivation

Be able to inject headers for not recording spans.

# Additional Notes


iunanua:
The problem exposed here is fixed after the implementation of the datadog Sampler.

It [returns](https://github.com/DataDog/dd-trace-rs/blob/main/dd-trace-sampling/src/datadog_sampler.rs#L487) `SamplingDecision::RecordOnly` instead of `SamplingDecision::Drop` for traces with drop priority, causing spans to be managed by `DatadogSpanProcessor::on_start` and therefore its propagation data to be stored correctly.

PR #35, opened 2025-06-13
feat(sampler): propagate the agent response back to the sampler to sample by agent rates

paullegranddc:
# What does this PR do?

This PR exposes a callback from the trace exporter into the sampler to return agent rates.

The agent rates is in a RwLock because it need to be periodically sampled.

This PR also depends on an unmerged branch of the trace exporter which needs to be updated once the linked PR is merged.
https://github.com/DataDog/libdatadog/pull/1104

# Motivation

# Additional Notes

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -32,7 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rust_version: ["1.81.0", "stable", "nightly-2024-12-16"]
+        rust_version: ["1.84.1", "stable", "nightly-2024-12-16"]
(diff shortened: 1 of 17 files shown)


bantonsson (approved):
Looks good, but it feels like we need to sort out the MSRV story with `libdatadog`.

paullegranddc:
>Looks good, but it feels like we need to sort out the MSRV story with libdatadog.

We agreed that libdatadog wouldn't upgrade MSRV for at least 6 month, at which point 1.84 will be a year old.
By then hopefully we will have customer data on which versions do they use and can define a policy

PR #37, opened 2025-06-24
feat(propagation): protect text_map_propagator with catch_panic!

iunanua:
# What does this PR do?

- introduce the `catch_panic` macro to wrap operations with `panic::catch_unwind` and protect them from panics.
- change otel `TextMapPropagator` to use the macro in `inject_context` and `extract_with_context`

# Motivation

Isolate customer code from possible panics produced by datadog's propagation logic.


iunanua:
@paullegranddc what do you think about this PR? it may be interesting or not worth it.

PR #40, opened 2025-06-26
feat(telemetry): app-started, app-heartbeat and logs

iunanua:
# What does this PR do?

- Add `language`, `telemetry_enabled`, `telemetry_log_collection_enabled` and `telemetry_heartbeat_interval` properties to `Config`
- Import `ddtelemetry` crate
- Provide the `Telemetry::init_telemetry` function to start a `TelemetryWorker` and start sending telemetry payloads like app-started, app-heartbeat
- Provide `Telemetry::add_log_error` and call it from `dd_error!` macro to send errors via telemetry. So each time `dd_error!` macro is called the message and a "stack trace" containing file and line number is sent.

# Notes
Telemetry is initialized when `DatadogSpanProcessor::set_resource` is executed in order to ensure all the different components use the same service name

# Motivation

It will be useful to have telemetry data in the dogfooding phase

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -149,7 +149,7 @@ jobs:
       - name: Cache
         uses: ./.github/actions/cache
       - name: Install dd-rust-license-tool
-        run: cargo install --git https://github.com/DataDog/rust-license-tool.git dd-rust-license-tool
+        run: which dd-rust-license-tool || cargo install --git https://github.com/DataDog/rust-license-tool.git dd-rust-license-tool
       # To ensure the dependency tree is identical between the CI runner and the
       # local license generation script (which also forces the 'git' protocol),
       # we explicitly set the crates.io protocol. Cargo's default can vary
--- a/README.md
+++ b/README.md
@@ -13,21 +13,22 @@ The `datadog-opentelemetry` crate provides an easy to use override for the rust 
 
 ```rust
 use std::time::Duration;
-use dd_trace::Config;
-use opentelemetry_sdk::trace::TracerProviderBuilder;
 
 fn main() {
     // This picks up env var configuration and other datadog configuration sources
-    let datadog_config = Config::builder().build();
-    let tracer_provider = datadog_opentelemetry::init_datadog(
-        datadog_config,
-        TracerProviderBuilder::default(),
-        None,
-    );
+    let tracer_provider = datadog_opentelemetry::tracing()
+        .init();
 
     // Your code
+    // Now use standard OpenTelemetry APIs
+    use opentelemetry::global;
+    use opentelemetry::trace::Tracer;
 
+    let tracer = global::tracer("my-service");
+    let span = tracer.start("my-operation");
+    // ... do work ...
 
+    // Shutdown the tracer to flush the remaining data
     tracer_provider.shutdown_with_timeout(Duration::from_secs(1)).expect("tracer shutdown error");
 }
 ```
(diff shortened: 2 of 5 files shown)


dd-trace/src/telemetry.rs
```diff
@@ -0,0 +1,265 @@
...
+use ddtelemetry::{
+    data::{self},
+    worker::{self, TelemetryWorkerHandle},
+};
+
+use crate::{dd_debug, dd_error, dd_info, Config};
+
+static INIT_TELEMETRY_LOCK: Mutex<()> = Mutex::new(());
+
+static TELEMETRY: OnceLock<Arc<Mutex<Telemetry>>> = OnceLock::new();
```
paullegranddc, line 18: I don't think you need both `Mutex` and `OnceLock` 🤔 

The pattern for `OnceLock` is usually

```rust
static TELEMETRY: OnceLock<Arc<Mutex<Telemetry>>> = OnceLock::new();

fn init_telemetry() {
  let Some(telemetry) = TELEMETRY. get_or_init(|| {
    // Create and start the handle here
    // This closure is only run once in the lifetime of the process
    match make_telemetry_worker() {
      Ok(handle) => Some(handle),
      Err(e) => { /* handle err */ ; None }, 
    }
  }) else {
    // there was an error and it has been logged already
    // return
  };
  // Do whatever you need with the handle here
}
```

dd-trace/src/telemetry.rs
```diff
@@ -0,0 +1,265 @@
...
+                if let Some(handle) = &telemetry.handle {
+                    handle.add_error_log(message.into(), stack).ok();
+                }
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+#[serial_test::serial]
```
paullegranddc, line 141: By the way, if you want to make this testable in parralel, you can do something like this

```rust
pub fn init_telemetry() {
  init_telemetry_inner(&TELEMETRY)
}

fn init_telemetry_inner(telemetry_cell: &OnceLock) {
  // do init using a parameter instead of a global
}

pub fn add_log() {
  add_log_inner(&TELEMETRY)
}

fn add_log_inner(telemetry: &OnceLock) {
}



// in tests each test gets it's own OnceLock instance
#[test]
fn test_init_telemetry() {
  let telemetry_cell = OnceLock::new();
  init_telemetry_inner(&telemetry_cell);
  send_log_inner(&telemetry_cell);
 }

```

dd-trace/src/telemetry.rs
```diff
@@ -0,0 +1,265 @@
...
+pub fn add_log_error<I: Into<String>>(message: I, stack: Option<String>) {
+    if let Some(telemetry) = TELEMETRY.get() {
+        if let Ok(telemetry) = telemetry.lock() {
+            if telemetry.enabled && telemetry.log_collection_enabled {
+                if let Some(handle) = &telemetry.handle {
+                    handle.add_error_log(message.into(), stack).ok();
+                }
+            }
+        }
+    }
```
paullegranddc, line 137: Nit but I prefer early return
```suggestion
    let Some(telemetry) = TELEMETRY.get() else { return; }
    let Ok(telemetry) = telemetry.lock() else { return; }
    if !telemetry.enabled || !telemetry.log_collection_enabled {
        return; 
    }
    let Some(handle) = &telemetry.handle else {
        return;
    }
    handle.add_error_log(message.into(), stack).ok();
```

dd-trace/src/telemetry.rs
```diff
@@ -0,0 +1,265 @@
...
+
+        return;
+    } else if !config.telemetry_enabled() {
+        dd_info!("Telemetry not enabled");
+        return;
+    }
+
+    let handle: Option<Box<dyn TelemetryHandle>> = if custom_handle.is_none() {
+        let mut builder = worker::TelemetryWorkerBuilder::new(
+            "127.0.0.1".to_string(), // FIXME
```
paullegranddc, line 78: This should be the `trace_agent_url` parameter from the config

dd-trace/src/telemetry.rs
```diff
@@ -0,0 +1,265 @@
...
+    worker::{self, TelemetryWorkerHandle},
+};
+
+use crate::{dd_debug, dd_error, dd_info, Config};
+
+static INIT_TELEMETRY_LOCK: Mutex<()> = Mutex::new(());
+
+static TELEMETRY: OnceLock<Arc<Mutex<Telemetry>>> = OnceLock::new();
+
+pub trait TelemetryHandle: Sync + Send + 'static {
```
paullegranddc, line 20: this trait doesn't need to be public

dd-trace/src/telemetry.rs
iunanua: done!

dd-trace/src/telemetry.rs
iunanua: Changed also

dd-trace/src/telemetry.rs
iunanua: As `init_telemetry` fn is public, `TelemetryHandle` must also be public, isn't it?

paullegranddc (approved):
LGTM

PR #49, opened 2025-07-09
chore: improve configuration & sample usability

scottgerring:
# What does this PR do?

**General sample usability**
Improve sample usability - makes them both cargo projects, and adds a README. Previously `simple_tracing` could be run with `cargo run --example` but propagator which requires 3rd party deps does not.

**Builder fluent-interface style** 
Whilst playing with the samples, I noticed you couldn't use the builders in the typical fluent fashion:
```rust
let config = dd_trace::Config::builder()
    .with_service("simple_tracing".to_string())
    .build();
```
i've changed this to consistently return `&mut self`, and have `build()` clone the config, per the [effective rust recommendation](https://effective-rust.com/builders.html).

**default for Config**
Removed the `impl Default` for `Config` so we don't confuse users outside, and added a  `default_config()` that we can use internally. 

**license tool caching**
Was missing `~/.cargo/.crates2.json`, which made the cache restore and subsequent cargo install of the license tool fail periodically; i've added comments on the review view.

# Motivation

A quest for great DX! 

# Additional Notes


LICENSE-3rdparty.yml
```diff
@@ -1,4 +1,4 @@
-root_name: datadog-opentelemetry, datadog-opentelemetry-mappings, dd-trace, dd-trace-propagation, dd-trace-sampling
+root_name: datadog-opentelemetry, datadog-opentelemetry-mappings, dd-trace, dd-trace-propagation, dd-trace-sampling, propagator
```
scottgerring, line 1: I don't understand how the license tool is working, but not including this causes a diff 🤷

scottgerring:
@paullegranddc this should be good to go!

LICENSE-3rdparty.yml
bantonsson: Yeah we should probably remove the `.yml` one and only have the `.csv` one.

bantonsson (approved):
Thanks for cleaning this up

PR #50, opened 2025-07-14
chore: upgrade to otel 0.30.0

scottgerring:
# What does this PR do?

# Motivation

# Additional Notes


paullegranddc (approved):

paullegranddc:
I added a change to still continue parsing the old attributes.
I think that since we're not an instrumentation, but a library consumming it we should probably be flexible in what attributes users can pass to us and accept use of deprecated attributes

PR #54, opened 2025-08-08
fix(telemetry): remove dynamic messages from stack_trace and review error logs

iunanua:
# What does this PR do?

Remove dynamic messages from stack_trace and review error log messages.

# Motivation

The inclusion of dynamic messages in the stack_trace is not correct

# Additional Notes

Should we define a format on how the error logs should be?

### DIFF
--- a/dd-trace/src/log.rs
+++ b/dd-trace/src/log.rs
@@ -140,7 +140,7 @@ pub fn print_log<I: Into<String>>(
         eprintln!("\x1b[91m{lvl}\x1b[0m {file}:{line} - {log}");
 
         if let Some(template) = template {
-            // there should be no dynamic messages in either log message or log stack_trace
+            // we should only send the template to telemetry to not leak sensitive information
             crate::telemetry::add_log_error(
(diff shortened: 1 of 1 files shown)


dd-trace/src/log.rs
```diff
@@ -139,10 +139,13 @@ pub fn print_log<I: Into<String>>(
     if lvl == crate::log::LevelFilter::Error {
         eprintln!("\x1b[91m{lvl}\x1b[0m {file}:{line} - {log}");
 
-        crate::telemetry::add_log_error(
-            template.unwrap_or(&log),
-            Some(format!("Error: {log}\n at {file}:{line}")),
-        );
+        if let Some(template) = template {
+            // there should be no dynamic messages in either log message or log stack_trace
```
bantonsson, line 143: I find this comment a bit confusing. I think it would be better if it said what we do and why, ie. `// we should only send the template to telemetry to not leak sensitive information`.

bantonsson (approved):

PR #56, opened 2025-08-12
fix: config consistency fixes

iunanua:
# What does this PR do?

- Ignore spans if `config.enabled()` is false
- Add `env` and `version` meta attributes when exporting after service has been resolved
- Mute one error sent via telemetry when `x-datadog-trace-id` header is not found (this is not related with the fix I can move it to another PR but I have noticed it when testing the other cases)

# Motivation

test_config_consistency::Test_Config_TraceEnabled
test_config_consistency::Test_Config_UnifiedServiceTagging

### DIFF
--- a/datadog-opentelemetry/src/ddtrace_transform.rs
+++ b/datadog-opentelemetry/src/ddtrace_transform.rs
@@ -49,11 +49,32 @@ pub fn otel_trace_chunk_to_dd_trace_chunk(
                 otel_resource,
             );
             otel_sampling_to_dd_sampling(trace_flags, &mut dd_span);
-            if dd_span.service == datadog_opentelemetry_mappings::DEFAULT_OTLP_SERVICE_NAME {
-                dd_span.service = BytesString::from_string(cfg.service().to_string());
-            }
+
+            add_config_metadata(&mut dd_span, cfg);
 
             dd_span
         })
         .collect()
+}
+
+fn add_config_metadata(dd_span: &mut DdSpan, cfg: &dd_trace::Config) {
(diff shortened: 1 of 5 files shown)


datadog-opentelemetry/src/ddtrace_transform.rs
```diff
@@ -49,11 +49,44 @@ pub fn otel_trace_chunk_to_dd_trace_chunk(
                 otel_resource,
             );
             otel_sampling_to_dd_sampling(trace_flags, &mut dd_span);
-            if dd_span.service == datadog_opentelemetry_mappings::DEFAULT_OTLP_SERVICE_NAME {
-                dd_span.service = BytesString::from_string(cfg.service().to_string());
-            }
+
+            add_config_metadata(&mut dd_span, cfg);
```
bantonsson, line 53: Is the work in `add_config_metadata` repeated for every span? Can't we do the cfg splitting and checking once, and just add things in the loop?

iunanua: Simple case implemented

dd-trace-propagation/src/datadog.rs
```diff
@@ -157,6 +157,8 @@ fn validate_tag_value(value: &str) -> bool {
 }
 
 pub fn extract(carrier: &dyn Extractor) -> Option<SpanContext> {
+    carrier.get(DATADOG_TRACE_ID_KEY)?;
```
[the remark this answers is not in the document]
bantonsson: Can't chalk this up to Monday morning, but I missed the `?` at the end of the line... 🤦🏼‍♂️

PR #67, opened 2025-08-29
feat(remote_config): make the remote config worker shutdown when the tracer stops

paullegranddc:
# What does this PR do?

* Split the RemoteConfigClient in two. Just the client to perform related to fetching and applying remote config changes, and a RemoteConfigClientWorker that spawns a thread to perform the fetch loop.
Whenever we create a worker, we get back a handle that can be used to trigger and wait for shutdown.
The worker also waits on a cancellation tokens, and signals back whenever it shuts down.

* Refactor a bit the client itself. While trying to understand the code, I found a few ways to make it easier to read.

* Add integration tests for remote config using the trace agent

* Refactor some of the logic related to the SpanProcessor shutdown. Now we first trigger both the exporter and rc worker shutdown, and then wait for them in order

# Motivation

Whenever we create a new Tracer, we also create a RC background worker. These workers are never stopped, and since they hold callbacks into the sampler, they keep quite a lot of memory alive forever which could cause memory leaks.
Also there are no integration tests using RC rules.

# Additional Notes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -667,7 +667,7 @@ checksum = "2a2330da5de22e8a3cb63252ce2abb30116bf5265e89c0e01bc17015ce30a476"
 [[package]]
 name = "data-pipeline"
 version = "20.0.0"
-source = "git+https://github.com/Datadog/libdatadog?rev=38ec3833177daeb3fb6d9fea501d918e5208e62a#38ec3833177daeb3fb6d9fea501d918e5208e62a"
+source = "git+https://github.com/Datadog/libdatadog?rev=0f88ff8acf5aff557fa9b786383b62b3f51d17a0#0f88ff8acf5aff557fa9b786383b62b3f51d17a0"
 dependencies = [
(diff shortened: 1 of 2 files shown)


dd-trace/src/configuration/remote_config.rs
```diff
@@ -236,27 +436,37 @@ struct SignedTargets {
...
-    agent_url: String,
+    agent_url: hyper::Uri,
     // HTTP client timeout configuration
     client_timeout: Duration,
     state: Arc<Mutex<ClientState>>,
     capabilities: ClientCapabilities,
     poll_interval: Duration,
     // Cache of successfully applied configurations
-    cached_target_files: Arc<Mutex<Vec<CachedTargetFile>>>,
+    cached_target_files: Vec<CachedTargetFile>,
```
bantonsson, line 451: Nice

bantonsson (approved):

iunanua (approved):

PR #73, opened 2025-09-05
fix(system-test): use ST main branch

iunanua:
# What does this PR do?

- Changes parametric job to use ST main branch
- Removes `_experimental_runs_on` parameter (`dd-trace-rs` is on the `APM Larger Runners` group now)

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -126,7 +126,7 @@ jobs:
   parametric:
     needs:
       - build-artifacts
-    uses: DataDog/system-tests/.github/workflows/run-parametric.yml@main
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -126,12 +126,10 @@ jobs:
   parametric:
     needs:
       - build-artifacts
-    uses: DataDog/system-tests/.github/workflows/run-parametric.yml@a675fcc0b1d582485c57c4ab0c30cf2f9eb217ea # igor/rust-the-germ
+    uses: DataDog/system-tests/.github/workflows/run-parametric.yml@main
```
bantonsson, line 129: I think that we should have a fixed hash here so we can move it and include new tests as we see fit. Having a moving parametric test target would also make supporting multiple versions hard (yes, we're not there yet).

PR #74, opened 2025-09-05
feat(telemetry): add config telemetry

iunanua:
# What does this PR do?

Changes all configurable properties of Config struct to allow tracking changes to their values and it the origin of those values.
At the moment there are two types:
- `ConfigItem<T>` for "immutable" properties
- `ConfigItemWithOverride<T>` for properties that change via RC or code 

# Motivation

Send config properties via telemetry and prepare the code for the RC system tests


dd-trace/src/configuration/configuration.rs
```diff
@@ -436,26 +670,26 @@ pub struct Config {
     language: &'static str,
 
     // # Service tagging
-    service: ServiceName,
-    env: Option<String>,
-    version: Option<String>,
+    service: ConfigItem<ServiceName>,
+    env: ConfigItem<OptionalString>,
```
iunanua, line 674: I had to create all these types (`OptionalString`, `GlobalTags`, `TracePropagationStyleList`) to implement `Display`

dd-trace/src/configuration/configuration.rs
```diff
@@ -657,6 +887,34 @@ impl Config {
         Self::builder_with_sources(&CompositeSource::default_sources())
     }
 
+    pub fn get_config_items(&self) -> Vec<Configuration> {
+        vec![
+            self.service.get_configuration(),
```
iunanua, line 892: Is there a way to automate this list? like a macro to iterate struct fields...

dd-trace/src/configuration/configuration.rs
```diff
@@ -214,25 +238,88 @@ impl<T: Clone> ConfigItem<T> {
...
+            ConfigSourceOrigin::Code
+        } else if self.env_value.is_some() {
+            ConfigSourceOrigin::EnvVar
+        } else {
+            ConfigSourceOrigin::Default
+        }
+    }
+
+    fn get_configuration(&self) -> Configuration {
+        Configuration {
```
iunanua, line 261: Configuration could be cached in a property of `ConfigItem`

dd-trace/src/configuration/configuration.rs
```diff
@@ -412,6 +600,52 @@ impl ServiceName {
...
 }
 
+impl Display for ServiceName {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.as_str())
+    }
+}
+
+#[derive(Clone, Debug, PartialEq)]
+enum OptionalString {
```
iunanua, line 610: ugly but I cannot `impl Display for Option<String>`

bantonsson:
This feels a bit convoluted, and I'm wondering if we don't have a _fixed_ set of config types, i.e. `Option<String>`, `Option<intX>`, `Vec<(String, String)`, and the like, where we could have an `Enum` instead of the `Display` for `T` in the `ConfigItem` and just implement `Display` for that `Enum`?

dd-trace/src/configuration/configuration.rs
iunanua: Approach changed

iunanua:
> This feels a bit convoluted
I have changed the approach. Instead of using `T: Display` I'm using a new `ConfigurationValueProvider` trait. 
It is automatically implemented by a macro for a lot of the types used in config properties and the rest are implemented manually.

dd-trace/src/configuration/configuration.rs
```diff
@@ -156,53 +181,56 @@ impl<T: Deref> Deref for ConfigItemRef<'_, T> {
     }
 }
 
+impl<T: std::fmt::Display> std::fmt::Display for ConfigItemRef<'_, T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        (**self).fmt(f)
+    }
+}
+
+trait ConfigurationValueProvider {
```
bantonsson, line 190: This needs documentation. The difference between `get_configuration_value` and `value` in `ConfigItem` is unclear from just reading the code.

dd-trace/src/configuration/configuration.rs
```diff
@@ -156,112 +181,325 @@ impl<T: Deref> Deref for ConfigItemRef<'_, T> {
...
+        } else {
+            self.config_item.source()
+        }
+    }
+
+    /// Replaces override value only if origin matches source_type
+    fn set_override_value(&self, value: T, source: ConfigSourceOrigin) {
+        if source == self.source_type {
+            self.override_value.store(Some(Arc::new(value)));
+        }
```
paullegranddc, line 357: If this is already called `set_override_value` why are we checking that the source is correct, and doing nothing if it's not?

I'm not sure we need the check, or if we need the check to protect against bugs it should return an error. Otherwise we're going to miss config updates

dd-trace/src/configuration/configuration.rs
```diff
@@ -673,62 +966,63 @@ impl Config {
         self.language_version.as_str()
     }
 
-    pub fn service(&self) -> &str {
-        self.service.as_str()
+    pub fn service(&self) -> String {
+        self.service.value().to_string()
```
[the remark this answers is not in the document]
iunanua: ok, I have recovered it. But still I cannot return a `&str` 
```rust
fn service(&self) -> &str {
    &*self.service.value()  // Error: temporary value dropped
}
```


and I have to handle the different `ConfigItemRef` cases.
```rust
pub fn service(&self) -> Cow<'_, str> {
    match self.service.value() {
        ConfigItemRef::Ref(t) => Cow::Borrowed(t.as_str()),
        ConfigItemRef::ArcRef(guard) => {
            Cow::Owned(guard.as_ref().unwrap().as_str().to_string())
        }
    }
}
```


Am I missing something?

dd-trace/src/configuration/configuration.rs
```diff
@@ -673,62 +966,63 @@ impl Config {
...
 
     pub fn dogstatsd_agent_url(&self) -> &Cow<'static, str> {
-        &self.dogstatsd_agent_url
+        self.dogstatsd_agent_url.value()
     }
 
     pub fn trace_sampling_rules(&self) -> impl Deref<Target = [SamplingRuleConfig]> + use<'_> {
-        self.trace_sampling_rules.value()
+        self.trace_sampling_rules.value().to_vec()
     }
```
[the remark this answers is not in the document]
iunanua: it is unnecessary

dd-trace/src/configuration/configuration.rs
iunanua: My idea was to allow only updates for the source for which `ConfigItemWithOverride` was created for, if a `ConfigItemWithOverride` is created for RC updates I don't want to update it with a Code source...

I could log a warn if the value is not updated or return a `Result` to indicate the caller if the update went OK or not

dd-trace/src/configuration/configuration.rs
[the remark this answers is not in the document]
paullegranddc: I think the idea is you return a `impl Deref<Target = str> + use<'_>`
The reason you cannot return a reference from  `ConfigItemRef` is that `ConfigItemRef::ArcRef` holds a guard and you cannot read the reference after it has been dropped. But since ConfigItemRef implements Deref you can return an opaque value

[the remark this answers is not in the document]
iunanua: ahhh true! :facepalm:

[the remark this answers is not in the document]
iunanua: and for `service_is_default()`, is there a way to avoid the match?
```rust
pub fn service_is_default(&self) -> bool {
    match self.service.value() {
        ConfigItemRef::Ref(t) => t.is_default(),
        ConfigItemRef::ArcRef(guard) => guard.as_ref().unwrap().is_default(),
    }
}
```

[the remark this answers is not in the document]
paullegranddc: I'm afraid not 🤔

paullegranddc (approved):

PR #81, opened 2025-09-15
feat(propagation): reduce inject allocations

iunanua:
# What does this PR do?

Use a single String buffer to avoid intermediate Vec allocation when injecting propagation tags.

### DIFF
--- a/dd-trace-propagation/src/context.rs
+++ b/dd-trace-propagation/src/context.rs
@@ -170,7 +170,10 @@ impl FromStr for Tracestate {
                                 None
                             } else {
                                 let mut parts = item.splitn(2, ':');
-                                Some((parts.next()?.to_string(), decode_tag_value(parts.next()?)))
+                                Some((
+                                    parts.next()?.to_string(),
+                                    decode_tag_value(parts.next()?).to_string(),
+                                ))
                             }
                         })
                         .collect(),
@@ -236,8 +239,12 @@ impl FromStr for Tracestate {
     }
 }
 
-fn decode_tag_value(value: &str) -> String {
-    value.replace('~', "=")
+fn decode_tag_value(value: &str) -> Cow<'_, str> {
+    if value.as_bytes().contains(&b'~') {
+        Cow::Owned(value.replace('~', "="))
+    } else {
+        Cow::Borrowed(value)
+    }
 }
 
 pub fn encode_tag_value(tag: &str) -> Cow<'_, str> {


dd-trace-propagation/src/datadog.rs
```diff
@@ -31,8 +31,12 @@ const DATADOG_PROPAGATION_ERROR_KEY: &str = "_dd.propagation_error";
 pub const DATADOG_LAST_PARENT_ID_KEY: &str = "_dd.parent_id";
 
 // TODO: get max_length from config: DD_TRACE_X_DATADOG_TAGS_MAX_LENGTH
+#[cfg(not(feature = "test-utils"))]
 pub const DD_TRACE_X_DATADOG_TAGS_MAX_LENGTH: usize = 512;
 
+#[cfg(feature = "test-utils")]
+pub const DD_TRACE_X_DATADOG_TAGS_MAX_LENGTH: usize = 51200;
```
bantonsson, line 38: Why?

iunanua: Removed!

dd-trace-propagation/src/tracecontext.rs
```diff
@@ -84,85 +84,117 @@ fn inject_traceparent(context: &SpanContext, carrier: &mut dyn Injector) {
...
+            if let Some(id) = context.tags.get(DATADOG_LAST_PARENT_ID_KEY) {
+                dd_parts.push_str(id);
             } else {
-                dd
+                dd_parts.push_str(&format!("{:016x}", context.span_id));
             }
-        })
-        .unwrap_or_default();
+        } else {
+            dd_parts.push_str(&format!("{:016x}", context.span_id));
```
paullegranddc, line 132: Could be `let _ = write!(&mut dd_parts, "{:016x}", context.span_id);` to save one allocation
(the `let _ =` is necessary because write! returns a Result, although we're sure it's going to be Ok)

dd-trace-propagation/src/tracecontext.rs
```diff
@@ -84,85 +84,117 @@ fn inject_traceparent(context: &SpanContext, carrier: &mut dyn Injector) {
...
+    dd_parts.push(':');
+    dd_parts.push_str(&priority.to_string());
 
-    tracestate_parts.push(format!("{TRACESTATE_SAMPLING_PRIORITY_KEY}:{priority}"));
-
-    if let Some(origin) = context.origin.as_ref().map(|origin| {
-        encode_tag_value(
+    // Build origin part if present
+    if let Some(origin) = context.origin.as_ref() {
+        let origin_encoded = encode_tag_value(
```
paullegranddc, line 98: `encode_tag_value` always copies the string, even if not modifications are done.
You could replace it with something like 
```rust
fn encode_tag_value(tag: &str) -> Cow<'_, str> {
  if tag.as_bytes().contains(b'=') {
    Cow::Owned(tag.replace('=', "~"))
  } else {
    Cow::Borrowed(tag)
  }
}
```

paullegranddc (approved):

PR #85, opened 2025-09-19
fix(remote_config): clear sampling rules on explicit null

iunanua:
# What does this PR do?

- Add a new deserializer to distinguish between missing `tracing_sampling_rules` field and `tracing_sampling_rules: null`
- Call `clear_remote_sampling_rules()` when null is received

# Motivation

# Additional Notes

### DIFF
--- a/dd-trace/src/configuration/remote_config.rs
+++ b/dd-trace/src/configuration/remote_config.rs
@@ -171,6 +171,20 @@ struct TargetFile {
     raw: String,
 }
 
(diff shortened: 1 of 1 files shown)


paullegranddc:
So if I understandand correctly, null will deserize to Some(Value::Null), which doesn't clear rules?
I don't understand why you add a custom deserializer to have null produce None, rather than adding an `else` condition line 860

PR #87, opened 2025-09-19
fix(clippy): rust stable upgraded to 1.90 and enabled new lints

paullegranddc:
# What does this PR do

Fix a clippy lint failing on main. It wasn't before but rust stable upgraded to 1.90 and enabled new lints, one of which is failing

### DIFF
--- a/dd-trace-propagation/src/context.rs
+++ b/dd-trace-propagation/src/context.rs
@@ -109,7 +109,7 @@ impl Tracestate {
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


bantonsson (approved):
Maybe we should run the lint with a fixed version?

paullegranddc:
> Maybe we should run the lint with a fixed version?

That wouldn't be a bad idea 😄

paullegranddc:
But I like the fact that we automatically stay compatible with stable without manual bumps

PR #92, opened 2025-09-26
feat(dd-trace-propagation): get rid of regex for replacing and validating invalid headers

paullegranddc:
# What does this PR do?

Replace regex usage in dd-trace-propagation with manual validation of character ranges and trace context headers.

# Motivation

This pops up in profiles, and hardcoded transformations are very likely to be much more performant for simple character byte range matching

# Additional Notes


dd-trace-propagation/src/datadog.rs
```diff
@@ -324,7 +321,7 @@ fn validate_sampling_decision(tags: &mut HashMap<String, String>) {
     let should_remove =
         tags.get(SAMPLING_DECISION_MAKER_TAG_KEY)
             .is_some_and(|sampling_decision| {
-                let is_invalid = !VALID_SAMPLING_DECISION_REGEX.is_match(sampling_decision);
+                let is_invalid = !matches!(sampling_decision.parse::<i8>().ok(), Some(-9..=-0));
```
bantonsson, line 324: Is single digit sampling decisions the _max_?

bantonsson (approved):
Nice! Do you have any numbers on this?

PR #95, opened 2025-10-01
perf(inject): write tracestate in one go

paullegranddc:
# What does this PR do?

Two optimizations:
* Don't use an intermediary buffer for the datadog part of the trace context and instead truncate if we have written too much
* Have a fast path for header normalization

# Motivation

# Additional Notes

### DIFF
--- a/dd-trace-propagation/src/tracecontext.rs
+++ b/dd-trace-propagation/src/tracecontext.rs
@@ -139,6 +139,44 @@ impl BufAppender<'_> {
     }
 }
 
+fn append_dd_propagation_tags(context: &InjectSpanContext, tags_buffer: &mut BufAppender) {
+    for (key, value) in context.tags.iter() {
+        let Some(key_suffix) = key.strip_prefix(DATADOG_PROPAGATION_TAG_PREFIX) else {
+            continue;
+        };
+
+        let t_key_suffix = replace_chars(
+            key_suffix,
+            |c| !matches!(c, b'!'..=b'+' | b'-'..=b'<' | b'>'..=b'~'),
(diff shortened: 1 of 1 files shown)


iunanua (approved):
This part is becoming difficult to read xD

dd-trace-propagation/src/tracecontext.rs
```diff
@@ -88,30 +89,66 @@ fn inject_traceparent(context: &SpanContext, carrier: &mut dyn Injector) {
...
+        "00-{:032x}-{:016x}-{flags}",
+        context.trace_id, context.span_id
+    );
 
     dd_debug!("Propagator (tracecontext): injecting traceparent: {traceparent}");
 
     carrier.set(TRACEPARENT_KEY, traceparent);
 }
 
+fn buff_appender(s: &mut String) -> BufAppender<'_> {
```
ekump, line 109: this function is `buff` but everything else is `buf`

dd-trace-propagation/src/lib.rs
```diff
@@ -181,6 +181,35 @@ impl DatadogCompositePropagator {
...
+            let mut concatenated: [u8; LEN] = [0; LEN];
+            let mut at = 0;
+            $(
+                crate::const_append($s.as_bytes(), &mut concatenated, at);
+                at += $s.len();
+            )*
+            let _ = at;
+            concatenated
+        };
+        unsafe { std::str::from_utf8_unchecked(&CONCATENATED) }
```
ekump, line 208: should there be a `//safety ...` comment for the macro?

PR #96, opened 2025-10-01
feat(propagation): Inject borrowed data

paullegranddc:
# What does this PR do?

Injection needs to read fields such as propagation tags and origin. Currently the SpanContext struct needs to have ownership of these which forces copying when injecting the span context.

This PR introduces and InjectSpanContext which just borrows the fields

# Motivation

# Additional Notes

### DIFF
--- a/dd-trace-propagation/benches/inject_benchmark.rs
+++ b/dd-trace-propagation/benches/inject_benchmark.rs
@@ -1,7 +1,7 @@
 // Copyright 2025-Present Datadog, Inc. https://www.datadoghq.com/
 // SPDX-License-Identifier: Apache-2.0
 
-use criterion::{black_box, criterion_group, criterion_main, BatchSize, Criterion};
+use criterion::{criterion_group, criterion_main, BatchSize, Criterion};
 use dd_trace::{
     configuration::TracePropagationStyle,
     sampling::{mechanism, priority},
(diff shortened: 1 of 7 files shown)


dd-trace-propagation/benches/inject_benchmark.rs
```diff
@@ -97,7 +109,7 @@ fn bench_datadog_only_inject<M: criterion::measurement::Measurement + Measuremen
         b.iter_batched(
             || (create_simple_span_context(), BenchInjector::new()),
             |(mut context, mut carrier)| {
-                propagator.inject(black_box(&mut context), black_box(&mut carrier))
+                propagator.inject(&mut span_context_to_inject(&mut context), &mut carrier)
```
bantonsson, line 112: Does removing these `black_box` calls change the performance of the benchmark?

bantonsson (approved):
Same nit `perf` instead of `feat`

dd-trace-propagation/benches/inject_benchmark.rs
paullegranddc: The black box did nothing because criterion already wraps inputs and outputs of the measured function in black_box

bantonsson: Interesting. Didn't know that.

PR #97, opened 2025-10-02
fix(ci): make the free disk space step faster

paullegranddc:
Free Disk Space is currently the longest step in the ubuntu test job, and lasts up to 8mn!! By not removing large packages, this is reduced to 4mn and increase the disk size by 4GB of packages, which should not be an issue

# What does this PR do?

# Motivation

# Additional Notes

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -27,7 +27,7 @@ jobs:
           android: true
           dotnet: true
           haskell: true
-          large-packages: true
+          large-packages: false
           docker-images: false
           swap-storage: true
       - name: Checkout sources


bantonsson (approved):
Nice.

I wonder if disabling `mandb` makes a difference (I know it's updating the db a number of times).

https://stackoverflow.com/a/71669201

paullegranddc:
> I wonder if disabling mandb makes a difference (I know it's updating the db a number of times).

It works apparently
https://github.com/jlumbroso/free-disk-space/issues/40

paullegranddc:
There are other - unmerged - optimizations that look very promising. I think the repo is probably unmaintained, or at least passively maintained so I think it might be worth forking it and integrating all of the improvements

https://github.com/jlumbroso/free-disk-space/pull/26

PR #101, opened 2025-10-23
feat(config_inversion): [APMAPI-1627] Centralized configuration registry basic implementation

vpellan:
# What does this PR do?

This implements the basis of centralized configuration registry (FKA as config inversion) by adding a script to generate the `supported_configurations.rs` file from the `supported-configurations.json` file, and use the resulting enums instead of strings for accessing env vars.

# Motivation

As dd-trace-rs is not used by customers yet, it is a great opportunity to ensure we correctly document each configuration in it.

# Additional Notes

This does NOT implement yet: aliases, deprecations, handling of type, default values and propertyKeys.

### DIFF
--- a/datadog-opentelemetry/src/span_exporter.rs
+++ b/datadog-opentelemetry/src/span_exporter.rs
@@ -52,13 +52,13 @@ struct BatchFullError {
 /// Error that can occur when the mutex was poisoned.
 ///
 /// The only way to handle it is to log and try to exit cleanly
-struct MutexPoisonnedError;
+struct MutexPoisonedError;
 
 #[derive(Debug, PartialEq, Eq)]
 enum SenderError {
     AlreadyShutdown,
     TimedOut,
-    MutexPoisonned,
+    MutexPoisoned,
     BatchFull(BatchFullError),
 }
 
@@ -181,6 +181,7 @@ impl DatadogExporter {
                 .set_language_version(config.language_version())
                 .set_service(&config.service())
                 .set_output_format(TraceExporterOutputFormat::V04)
+                .enable_health_metrics()
                 .enable_agent_rates_payload_version();
 
             if config.trace_stats_computation_enabled() {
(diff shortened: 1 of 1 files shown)


supported-configurations.json
```diff
@@ -0,0 +1,268 @@
...
+  "version": 2,
+  "supportedConfigurations": {
+    "DD_AGENT_HOST": [
+      {
+        "version": "A",
+        "type": "string",
+        "default_value": "localhost",
+        "propertyKeys": [
+          "agent_host"
+        ]
```
paullegranddc: I don't think either `type` or `propertyKeys` are going to be used to generate code (or I don't see how we would do it?). Is it a good idea to define it here?

datadog-opentelemetry/src/sampler.rs
```diff
@@ -205,6 +205,7 @@ mod tests {
         );
 
         // Clean up environment
+        #[allow(clippy::disallowed_methods)]
```
paullegranddc: You can just delete the line 209, the remove_var is a leftover from previous code we forgot to cleanup

vpellan:
Hey ! Thanks for the review, I've added CI checks in 098bfc02b487c5e741df55a91bf8f59e6364a1e8 ! Could you also review them ? Thanks !

.github/workflows/configuration.yaml
```diff
@@ -0,0 +1,44 @@
...
+            if (sorted_config_keys === config_keys) process.exit(0);
+            if (sorted_config_keys == null || config_keys == null) process.exit(1);
+            if (sorted_config_keys.length !== config_keys.length) process.exit(1);
+
+            for (var i = 0; i < sorted_config_keys.length; ++i) {
+              if (sorted_config_keys[i] !== config_keys[i]) {
+                console.log(`${sorted_config_keys[i]} !== ${config_keys[i]}`);
+                process.exit(1);
+              }
+            }
```
paullegranddc: why not do the check as part of the `local_config_map_generate.py` script?
I feel like it's probably easier to have all of the logic in one place

vpellan: I did that in case devs manually modify both supported-configurations.json file and supported_configurations.rs without ordering the keys

supported-configurations.json
```diff
@@ -0,0 +1,268 @@
...
+        "type": "string",
+        "default_value": "",
+        "propertyKeys": [
+          "version"
+        ]
+      }
+    ]
+  },
+  "aliases": {},
+  "deprecations": {}
```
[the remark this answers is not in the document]
vpellan: Thanks! Removed in c47cd7b3fcc813828f423bff62ebf4c94a631533

PR #105, opened 2025-11-03
feat(span-exporter): add buffer full drop metrics

paullegranddc:
# What does this PR do?

Add exporter queue metrics:
* Spans dropped because the queue is full
* Move the spans enqueued metrics to the queue
* Fix the spans finished metric name to actually be "spans_finished" instead of "spans_dropped"

This PR also renames "stats" to "metrics" in the registry metrics so we use the same lexicon everywhere

# Motivation

# Additional Notes

### DIFF
--- a/dd-trace/src/telemetry.rs
+++ b/dd-trace/src/telemetry.rs
@@ -93,8 +93,8 @@ telemetry_metrics!(
     SpansFinished => ("spans_finished", Tracers, Count, []),
     SpansEnqueuedForSerialization => ("spans_enqueued_for_serialization", Tracers, Count, []),
     SpansDroppedBufferFull => ("spans_dropped", Tracers, Count, ["reason" => "overfull_buffer"]),
-    TraceSegmentsCreated => ("trace_chunks_created", Tracers, Count, []),
-    TraceSegmentsClosed => ("trace_chunks_closed", Tracers, Count, []),
+    TraceSegmentsCreated => ("trace_segments_created", Tracers, Count, []),
+    TraceSegmentsClosed => ("trace_segments_closed", Tracers, Count, []),
 );
 
 trait TelemetryHandle: Sync + Send + 'static + Any {


iunanua:
A couple of questions (I should have done them in the previous PR):
- Do we need `TelemetryMetricsCollector` and `TraceRegistryStats`? why don't call directly `telemetry::add_point()` where the metric value is get? for example in InnerTraceRegistry, instead of 
```rust
self.stats.trace_segments_created += 1;
self.stats.spans_created += 1;
```
call directly to
```rust
telemetry::add_point(1, TraceSegmentsCreated)
telemetry::add_point(1, SpansCreated)
```
- at the moment, each call to `telemetry::add_point` locks `Telemetry` object, checks enabled flag, gets the handle, etc. `TelemetryMetricsCollector::emit_metrics` calls `telemetry::add_point` 5 times in a row. Wouldn't be better to lock only once and add all the points at once?

paullegranddc:
> Do we need TelemetryMetricsCollector and TraceRegistryStats?

I did it for multiple reason:
* Easier testing
* I figured that it might be slightly more performant. Sending one message per event seems a bit heavy and the telemetry handle can drop metrics if we send too many points at once
* More of a personal preference, but it makes the code less coupled if we don't have calls depending on globals peppered all over the code

> Wouldn't be better to lock only once and add all the points at once?

Good remark! I added an API to submit a batch of points

dd-trace/src/telemetry.rs
```diff
@@ -84,11 +89,12 @@ macro_rules! telemetry_metrics {
...
-    SpansFinished => ("spans_dropped", Tracers, Count),
-    SpansEnqueuedForSerialization => ("spans_enqueued_for_serialization", Tracers, Count),
-    TraceSegmentsCreated => ("trace_chunks_created", Tracers, Count),
-    TraceSegmentsClosed => ("trace_chunks_closed", Tracers, Count),
+    SpansCreated => ("spans_created", Tracers, Count, []),
+    SpansFinished => ("spans_finished", Tracers, Count, []),
+    SpansEnqueuedForSerialization => ("spans_enqueued_for_serialization", Tracers, Count, []),
+    SpansDroppedBufferFull => ("spans_dropped", Tracers, Count, ["reason" => "overfull_buffer"]),
+    TraceSegmentsCreated => ("trace_chunks_created", Tracers, Count, []),
+    TraceSegmentsClosed => ("trace_chunks_closed", Tracers, Count, []),
```
iunanua: Are `trace_chunks_created` and `trace_chunks_closed` metrics defined in https://github.com/DataDog/dd-go/blob/prod/trace/apps/tracer-telemetry-intake/telemetry-metrics/static/common_metrics.json?

(`trace_segments_created` and `trace_segments_closed` are)

iunanua (approved):

PR #107, opened 2025-11-06
feat(span_processor): add debug mode for abandoned spans

paullegranddc:
# Motivation

This feature is inspired from the go tracer https://github.com/DataDog/dd-trace-go/pull/2188

It's hard to understand and locate spans that are never finished. This PR adds a debug mode to the tracer that will track the age and root span name of traces, and
* log warnings from time to time if the traces are older than some amount of time.
* log warning if some traces are still open during tracer shutdown

# Implementation

This feature should not add any cost to the tracer if it is not enabled, but requires storing some extra data associated with each trace.

In order to not use any extra memory if the feature is not enabled, I track it in an additional registry which is only used in debug mode.

This registry tracks the root span name, and the age of the trace.

The debug mode is controlled by 2 new configurations:
* DD_TRACE_DEBUG_OPEN_SPANS
* DD_TRACE_OPEN_SPAN_TIMEOUT

In order to test this feature correctly, I added extra code to intercept and store logs during integration tests. This is done through a thread local Logger, which is overridden and propagated locally during tests.

Everything is hidden behind the test-utils feature and should thus be zero cost

### DIFF
--- a/datadog-opentelemetry/src/abandoned_traces.rs
+++ b/datadog-opentelemetry/src/abandoned_traces.rs
@@ -163,7 +163,6 @@ impl InnerAbandonedTracesRegistry {
     }
 
     fn finish_span(&mut self, trace_id: [u8; 16], name: &str) {
-        dbg!("finish span");
         let Entry::Occupied(mut e) = self.traces.entry(trace_id) else {
             return;
         };
@@ -177,7 +176,6 @@ impl InnerAbandonedTracesRegistry {
         {
             trace.open_span_names.remove(name);
         };
-        dbg!(name, &trace.open_span_names);
         trace.open_spans -= 1;
         if trace.open_spans == 0 {
             e.remove();


supported-configurations.json
```diff
@@ -130,6 +130,22 @@
...
     ],
+    "DD_TRACE_DEBUG_OPEN_SPANS": [
+      {
+        "version": "A",
+        "type": "boolean",
+        "default_value": "false",
+        "propertyKeys": ["trace_debug_open_spans"]
+      }
+    ],
+    "DD_TRACE_OPEN_SPAN_TIMEOUT": [
```
bantonsson: This one is in the wrong alphabetical order. I also think it's weird that it's not named in the same way as `DD_TRACE_DEBUG_OPEN_SPANS`, since it's related. Naming it `DD_TRACE_DEBUG_OPEN_SPANS_TIMEOUT` would make more sense.

PR #109, opened 2025-11-12
chore(otel): bump compatible otel version to 0.31

paullegranddc:
# What does this PR do?

# Motivation

# Additional Notes

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -126,7 +126,7 @@ jobs:
   parametric:
     needs:
       - build-artifacts
-    uses: DataDog/system-tests/.github/workflows/run-parametric.yml@65bd7a6cd426f30bcabab6616fecdac7a882654e  # main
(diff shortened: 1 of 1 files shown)


datadog-opentelemetry/Cargo.toml
```diff
@@ -43,7 +43,7 @@ datadog-opentelemetry = { path = ".", features = ["test-utils"] }
 
 # Testing the tracing API
 tracing = { version = "0.1.4" }
-tracing-opentelemetry = { version = "0.31" }
+tracing-opentelemetry = { version = "0.32" }
```
bantonsson: Shouldn't this be `0.32.0`?

bantonsson: The way I read that, yes, since otherwise we would only get `0.31.1` an not `0.31.2`. But who knows 🤷🏼‍♂️

PR #114, opened 2025-11-19
chore(versioning): add release job

hoolioh:
# What does this PR do?

Add a publish workflow which will be triggered by tag. The tag must have the follwing format: `<crate_name>-vx.y.z`. The job is part of the `publish` environment which means it will be held until a member of `apm-rust` approves it. Upon approval, the job will perform some sanity checks on the crate and then publish it to crates.io.

### DIFF
--- a/scripts/publish-crate.sh
+++ b/scripts/publish-crate.sh
@@ -127,7 +127,7 @@ sort_tags() {
     local i=1
     for tag in "${final_sorted[@]}"; do
         echo -e "${BLUE}  $i. $tag${NC}" >&2
-        ((i++))
+        ((++i))
     done
     echo "" >&2
     
@@ -334,12 +334,12 @@ check_publication_status() {
     
     for tag in "${sorted_tags[@]}"; do
         if check_crate_only "$tag"; then
-            ((published++))
+            ((++published))
         else
             if [[ "$tag" =~ ^(.+)-v([0-9]+\.[0-9]+\.[0-9]+)$ ]]; then
-                ((not_published++))
+                ((++not_published))
             else
-                ((failed++))
+                ((++failed))
             fi
         fi
     done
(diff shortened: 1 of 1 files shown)


scripts/publish-crate.sh
```diff
@@ -0,0 +1,445 @@
...
+        publish_cmd="$publish_cmd --dry-run"
+        echo -e "${BLUE}[DRY RUN]${NC} $publish_cmd" >&2
+    fi
+    
+    if $publish_cmd; then
+        echo -e "${GREEN}✓ Published $crate_name v$crate_version successfully${NC}" >&2
+        echo "" >&2
+
+        if [ "$dry_run" != "true" ]; then
+            set_ownership "$crate_name" "$LIBDATADOG_OWNERS" "$token"
```
ekump: Does it matter if `set_ownership` fails? If so, I think this will report success even if `set_ownership` does fail. 

Would this report failure correctly? 

```bash
  if [ "$dry_run" != "true" ]; then
      if ! set_ownership "$crate_name" "$LIBDATADOG_OWNERS" "$token"; then
          echo -e "${RED}❌ Failed to set ownership${NC}" >&2
          return 1
      fi
  fi
  return 0
```

ekump (approved):
A couple of potential issues, LGTM if they aren't legitimate.

scripts/publish-crate.sh
```diff
@@ -0,0 +1,445 @@
...
+    
+    local sorted_tags
+    sorted_tags=($(sort_tags "${tags[@]}"))
+    
+    echo "" >&2
+    
+    # Check-only mode: just check publication status
+    if [ "$check_only" = true ]; then
+        local ret=check_publication_status "${sorted_tags[@]}"
+        exit ${ret}
```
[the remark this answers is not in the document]
hoolioh: you're completely right, switching between languages mess with my head :)

scripts/publish-crate.sh
hoolioh: Good catch

iunanua (approved):

PR #121, opened 2025-11-25
feat(api): refactor public api visibility

paullegranddc:
This PR breaks backward compatibility since we remove previously exported items

# Changes
* Mark modules we use in benchmarks as pub only if we have the test-utils feature.
* Re-export log and configuration publicly modules from the top level to hide module structure
* Mark remote config update related function as pub(crate) since they should not be called by users
* Remove the deprecated "init_datadog" method to create the tracer

# Motivation

We should be very deliberate in what APIs we want to expose. Exposing more than necessary will lead customers to rely on implementation details. For this purpose, we should try to not expose the current module structure of datadog-opentelemetry and put items at their "natural" import paths.

To explore what function is visible, "cargo doc" is a great tool

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -130,10 +130,11 @@ jobs:
   parametric:
     needs:
       - build-artifacts
-    uses: DataDog/system-tests/.github/workflows/run-parametric.yml@fe54df1ab9bb126ffb337b72bb9a2793d2466bf5  # main
+    uses: DataDog/system-tests/.github/workflows/run-parametric.yml@8a65f66d74ca77f34e7e5a2f737e963ce9cec563  # main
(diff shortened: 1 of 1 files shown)


datadog-opentelemetry/src/lib.rs
```diff
@@ -140,10 +141,27 @@
 //! * [`tracing-opentelemetry`](https://docs.rs/tracing-opentelemetry/0.32.0/tracing_opentelemetry/)
 //!   version: 0.32
 
+// Public re-exports
+pub use core::configuration;
```
iunanua: what if we re-export `configuration::Config` to avoid breaking the STs?

```suggestion
pub use configuration::Config;
pub use core::configuration;
```

PR #125, opened 2025-11-27
chore(ci): improve JUnit upload speed by using the standalone binary

juan-fernandez:
# What does this PR do?

Use `datadog-ci`'s standalone binaries to upload JUnit XML reports to Test Optimization, rather than using `datadog/junit-upload-github-action`, which uses `npx` under the hood, which can be slow. 

The version is pinned to 4.2.2 and I've added the checksum verification. See binaries and their checksums in https://github.com/DataDog/datadog-ci/releases/tag/v4.2.2 

# Motivation

The JUnit XML upload step can be slow, especially in windows machines. See example: https://github.com/DataDog/dd-trace-rs/actions/runs/19742962553/job/56571081346

Screenshot 2025-11-28 at 13 07 45

It is way faster now: https://github.com/DataDog/dd-trace-rs/actions/runs/19763254714/job/56629936833

Screenshot 2025-11-28 at 13 08 48

From ~4 min to ~5 seconds.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -75,54 +75,45 @@ jobs:
           report_paths: "target/nextest/ci/junit.xml"
           check_name: "[${{ matrix.platform }}:${{ matrix.rust_version }}] test report"
           include_passed: true
-      - name: Download datadog-ci binary
-        if: success() || failure()
-        shell: bash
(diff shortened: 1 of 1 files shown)


bantonsson (approved):
Thank you

PR #127, opened 2025-12-02
feat(otel): metrics support

mabdinur:
## What does this PR do?

Adds OpenTelemetry Metrics SDK support with OTLP transport to `datadog-opentelemetry`. Metrics can be exported via gRPC (default) or HTTP/protobuf, enabling Rust applications to export OpenTelemetry metrics to Datadog using the standard OpenTelemetry Metrics API.

### Key Features

- **OTLP Transport**: Supports gRPC (default, port 4317) and HTTP/protobuf (port 4318) protocols
- **Configuration API**: Full support for OTLP metrics configuration via `ConfigBuilder`, including `set_otel_resource_attributes()`
- **Resource Attributes**: Support for `OTEL_RESOURCE_ATTRIBUTES` environment variable
- **Temporality Control**: Configurable Delta/Cumulative temporality preference
- **Graceful Degradation**: HTTP/JSON protocol gracefully degrades with a warning (no-op provider)

## Motivation

Enables Rust applications to export OpenTelemetry metrics to Datadog using the standard OpenTelemetry Metrics API, aligning with support in other languages and providing a consistent observability experience across the Datadog ecosystem.

## Additional Notes

- **Metrics are opt-in** via `DD_METRICS_OTEL_ENABLED=true`

**Feature gating:**
- **Default**: `cargo add datadog-opentelemetry` → gRPC transport
- **HTTP-only**: `cargo add datadog-opentelemetry --no-default-features --features metrics-http` → HTTP/protobuf without gRPC dependencies
- **Both**: `cargo add datadog-opentelemetry --features metrics-grpc,metrics-http` → enables both protocols

**Testing:**
- Implementation is fully tested here: https://github.com/DataDog/system-tests/pull/5863

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1068,6 +1068,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "469fb0b9cefa57e3ef31275ee7cacb78f2fdca44e4765491884a2b119d4eb130"
 
 [[package]]
+name = "iri-string"
+version = "0.7.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c91338f0783edbd6195decb37bae672fd3b165faffb89bf7b9e6942f8b1a731a"
+dependencies = [
+ "memchr",
+ "serde",
+]
+
+[[package]]
 name = "is-terminal"
 version = "0.4.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1468,11 +1478,8 @@ dependencies = [
  "async-trait",
  "bytes",
  "http",
- "http-body-util",
- "hyper",
- "hyper-util",
  "opentelemetry",
- "tokio",
+ "reqwest",
 ]
 
 [[package]]
@@ -1487,6 +1494,7 @@ dependencies = [
  "opentelemetry-proto",
  "opentelemetry_sdk",
  "prost 0.14.1",
+ "reqwest",
  "thiserror 2.0.14",
  "tokio",
  "tonic",
@@ -1928,6 +1936,40 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2b15c43186be67a4fd63bee50d0303afffcef381492ebe2c5d87f324e1b8815c"
 
 [[package]]
+name = "reqwest"
+version = "0.12.28"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "eddd3ca559203180a307f12d114c268abf583f59b03cb906fd0b3ff8646c1147"
+dependencies = [
+ "base64 0.22.1",
+ "bytes",
+ "futures-channel",
+ "futures-core",
+ "futures-util",
+ "http",
+ "http-body",
+ "http-body-util",
+ "hyper",
+ "hyper-util",
+ "js-sys",
+ "log",
+ "percent-encoding",
+ "pin-project-lite",
+ "serde",
+ "serde_json",
+ "serde_urlencoded",
+ "sync_wrapper",
+ "tokio",
+ "tower",
+ "tower-http",
+ "tower-service",
+ "url",
+ "wasm-bindgen",
+ "wasm-bindgen-futures",
+ "web-sys",
+]
+
+[[package]]
 name = "rmp"
 version = "0.8.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2072,6 +2114,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "serde_urlencoded"
+version = "0.7.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d3491c14715ca2294c4d6a88f15e84739788c1d030eed8c110436aafdaa2f3fd"
+dependencies = [
+ "form_urlencoded",
+ "itoa",
+ "ryu",
+ "serde",
+]
+
+[[package]]
 name = "sha1"
 version = "0.10.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2187,6 +2241,9 @@ name = "sync_wrapper"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0bf256ce5efdfa370213c1dabab5935a12e49f2c58d15e9eac2870d3b4f27263"
+dependencies = [
+ "futures-core",
+]
 
 [[package]]
 name = "synstructure"
@@ -2453,6 +2510,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "tower-http"
+version = "0.6.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d4e6559d53cc268e5031cd8429d05415bc4cb4aefc4aa5d6cc35fbf5b924a1f8"
+dependencies = [
+ "bitflags",
+ "bytes",
+ "futures-util",
+ "http",
+ "http-body",
+ "iri-string",
+ "pin-project-lite",
+ "tower",
+ "tower-layer",
+ "tower-service",
+]
+
+[[package]]
 name = "tower-layer"
 version = "0.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2685,6 +2760,19 @@ dependencies = [
  "quote",
  "syn",
  "wasm-bindgen-shared",
+]
+
+[[package]]
+name = "wasm-bindgen-futures"
+version = "0.4.50"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "555d470ec0bc3bb57890405e5d4322cc9ea83cebb085523ced7be4144dac1e61"
+dependencies = [
+ "cfg-if",
+ "js-sys",
+ "once_cell",
+ "wasm-bindgen",
+ "web-sys",
 ]
 
 [[package]]


mabdinur:
Shared tests covering edge cases: https://github.com/DataDog/system-tests/pull/5863

datadog-opentelemetry/src/core/configuration/configuration.rs
```diff
@@ -1011,6 +1077,16 @@ impl Config {
             &self.remote_config_enabled,
             &self.remote_config_poll_interval,
             &self.datadog_tags_max_length,
+            &self.otlp_endpoint,
```
iunanua: is metrics_otel_enabled missing?
```suggestion
            &self.metrics_otel_enabled,
            &self.otlp_endpoint,
```

datadog-opentelemetry/src/lib.rs
```diff
@@ -477,3 +513,54 @@ pub fn make_test_tracer(shared_config: Arc<Config>) -> (SdkTracerProvider, Datad
...
+        self
+    }
+
+    pub fn init(self) -> Result<SdkMeterProvider, String> {
+        let (meter_provider, _) = self.init_local()?;
+        opentelemetry::global::set_meter_provider(meter_provider.clone());
+        Ok(meter_provider)
+    }
+
+    pub fn init_local(self) -> Result<(SdkMeterProvider, ()), String> {
```
iunanua: Is the tuple needed?
```suggestion
    pub fn init_local(self) -> Result<SdkMeterProvider, String> {
```

datadog-opentelemetry/Cargo.toml
```diff
@@ -76,7 +79,10 @@ tracing-subscriber = { version = "0.3" }
 criterion = "0.5.1"
 
 [features]
-test-utils = ["libdd-data-pipeline/test-utils", "criterion"]
+default = ["metrics-grpc"]
```
iunanua: I think we didn't want default features, wdyt @bantonsson?

Cargo.lock
paullegranddc: The Cargo lock file is somehow not up to date when I pull this branch.
You probably need to do a `cargo check` and commit the resulting lockfile

datadog-opentelemetry/src/lib.rs
```diff
@@ -477,3 +513,54 @@ pub fn make_test_tracer(shared_config: Arc<Config>) -> (SdkTracerProvider, Datad
...
+        let config = self.config.unwrap_or_else(|| Config::builder().build());
+        let meter_provider = crate::metrics_reader::create_meter_provider(
+            Arc::new(config),
+            self.resource,
+            self.export_interval,
+        )?;
+        Ok((meter_provider, ()))
+    }
+}
```
paullegranddc: I don't think those methods should ever return errors.
I'd prefer if we returned an no-op SdkMeterProvider (with no exporter) and logged a scary log, rather than customers having to think about what if the meter provider is missing.

Also in general, return `Result<T, String>` feels weird... I don't think it's completely invalid, but I'd prefer if you used `anyhow::Result<T>` instead.
The error is also typed erased and can be a just a string, but it also captures the backtrace and you can wrap errors more efficiently.

datadog-opentelemetry/src/metrics_exporter.rs
```diff
@@ -0,0 +1,81 @@
...
+pub fn get_otlp_metrics_timeout(config: &Config) -> u32 {
+    let timeout = config.otlp_metrics_timeout();
+    if timeout != 0 {
+        return timeout;
+    }
+    config.otlp_timeout()
+}
+
+pub fn get_metric_export_interval_ms(config: &Config) -> u32 {
+    config.metric_export_interval()
```
paullegranddc: A single line function not adding anything?

datadog-opentelemetry/src/lib.rs
```diff
@@ -165,10 +198,13 @@ pub(crate) mod propagation;
 pub(crate) mod sampling;
 
 mod ddtrace_transform;
+pub mod metrics_exporter;
+pub mod metrics_reader;
```
paullegranddc: Why are these modules publicly exposed?
The only item that  should probably be public is the `OtlpProtocol` enum, if we expose it in the config

datadog-opentelemetry/src/core/configuration/configuration.rs
```diff
@@ -780,6 +780,12 @@ pub struct Config {
     /// A list of default tags to be added to every span
     /// If DD_ENV or DD_VERSION is used, it overrides any env or version tag defined in DD_TAGS
     global_tags: ConfigItem<Vec<(String, String)>>,
+    /// OTEL resource attributes parsed from OTEL_RESOURCE_ATTRIBUTES env var
+    otel_resource_attributes: ConfigItem<Vec<(String, String)>>,
+    /// OTEL metrics exporter type
+    otel_metrics_exporter: ConfigItem<Cow<'static, str>>,
+    /// OTEL metrics temporality preference
+    otel_metrics_temporality_preference: ConfigItem<Cow<'static, str>>,
```
paullegranddc: same remark about type conversion with the temporality preference, I think it's better to do validation and cast to an enum here.

The manual `set_otel_metrics_temporality_preference` method should take an enum type instead of a string for instance

datadog-opentelemetry/src/metrics_reader.rs
```diff
@@ -0,0 +1,229 @@
...
+            Duration::from_millis(get_metric_export_interval_ms(&config) as u64)
+        });
+
+        let telemetry_exporter = TelemetryTrackingExporter::new(exporter, protocol);
+
+        let reader = opentelemetry_sdk::metrics::PeriodicReader::builder(telemetry_exporter)
+            .with_interval(interval)
+            .build();
+
+        let mut resource_attrs: Vec<opentelemetry::KeyValue> = Vec::new();
```
paullegranddc: Could you isolate all dd config -> otel resource manipulation to a separate function? I think it'd be easier to read

datadog-opentelemetry/src/telemetry_metrics_exporter.rs
```diff
@@ -0,0 +1,88 @@
...
+                OtelMetricsExportFailuresGrpc,
+            ),
+            OtlpProtocol::HttpProtobuf | OtlpProtocol::HttpJson => (
+                OtelMetricsExportAttemptsHttp,
+                OtelMetricsExportSuccessesHttp,
+                OtelMetricsExportFailuresHttp,
+            ),
+        };
+
+        telemetry::add_point(1.0, attempts);
```
paullegranddc: Is it fine to add one telemetry point per point exported? Or maybe exports are not that frequent?

datadog-opentelemetry/tests/integration_tests/metrics.rs
```diff
@@ -0,0 +1,232 @@
...
+            result.observe(15, &[KeyValue::new("key6", "value6")]);
+        })
+        .build();
+}
+
+#[tokio::test]
+async fn test_metrics_export_grpc() {
+    const SESSION_NAME: &str = "opentelemetry_api/test_metrics_grpc";
+
+    let test_agent = setup_test_agent(SESSION_NAME).await;
```
paullegranddc: The test agent version/client we use doesn't have the otlp endppints yet.
The support has been merged in libdatadog but we need to bump it here (we won't need a new release, just update to a git hash probably)

datadog-opentelemetry/src/lib.rs
mabdinur: Good catch. I will make these modules internal to the crate. They should not be exposed to users

datadog-opentelemetry/src/core/configuration/configuration.rs
mabdinur: We were actually missing the `set_otel_metrics_temporality_preference` method. I added it to the programatic api for configurations.

datadog-opentelemetry/src/metrics_reader.rs
mabdinur: I moved this logic to it's own method but I kept this logic in the metric reader (since this is the only place it's currently used). 

In a follow up PR I can move this logic to it's own module. Datadog SDKs should support the `OTEL_RESOURCE_ATTRIBUTES` configuration for tracing (and eventually logs). Here's a doc with all the configurations we need to support: https://docs.google.com/document/d/1gtdjhsMSIvuTnM9E4EreI7oZ-bAUoAxkhXX0-fBvdDY/edit?tab=t.0#heading=h.rnd972k0hiye

datadog-opentelemetry/src/telemetry_metrics_exporter.rs
mabdinur: We are adding one telemetry point per export. By default we will export metrics every 10 seconds, so this should be fine for most use cases.

datadog-opentelemetry/src/metrics_exporter.rs
```diff
@@ -0,0 +1,81 @@
...
+
+    let endpoint = config.otlp_endpoint();
+    if !endpoint.is_empty() {
+        return Ok(endpoint.to_string());
+    }
+
+    let agent_url = config.trace_agent_url();
+    let url = agent_url
+        .parse::<hyper::http::Uri>()
+        .map_err(|e| format!("Invalid agent URL: {e}"))?;
```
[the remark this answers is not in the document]
mabdinur: Nope. I updated this logic. At this point we can assume we have a valid url.

PR #133, opened 2025-12-11
fix(rc): parse the rc path properly

bantonsson:
# What does this PR do?

Fixes the parsing of the Remote Config Path.

# Motivation

The path in remote configuration requests for sampling rules failed to be parsed, and logged an error. 

# Additional Notes

Has been tested in staging with the UI fixes necessary to do trigger the RC updates.

### DIFF
--- a/datadog-opentelemetry/src/core/configuration/remote_config.rs
+++ b/datadog-opentelemetry/src/core/configuration/remote_config.rs
@@ -631,7 +631,8 @@ impl RemoteConfigClient {
 
             for file in target_files {
                 // Extract product and config_id from path to determine which handler to use
-                // Path format is like "datadog/2/{PRODUCT}/{config_id}/config"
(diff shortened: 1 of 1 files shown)


datadog-opentelemetry/src/core/configuration/remote_config.rs
```diff
@@ -857,18 +858,25 @@ impl ProductRegistry {
     }
 }
 
-/// Extract product name and id from remote config path
-/// Path format is: datadog/2/{PRODUCT}/{config_id}/config
+/// Extract product and id from remote config path
+/// Path format is: ^(datadog/\d+|employee)/[^/]+/[^/]+/[^/]+$
```
iunanua: would be interesting to add a test?

bantonsson: I'll do a follow up PR later with more tests and even better parsing 😬

PR #140, opened 2026-01-08
chore: remove warning and clippy suggestion

iunanua:
# What does this PR do?

Remove warning and clippy suggestion that idk how they got in.


iunanua:
> We still have a protection rule on `master` and the commit message doesn't match semantic commits, so the merge queue won't work.

I keep falling into the trap :S

PR #147, opened 2026-02-09
refactor(sampling): move opentelemetry code behind traits

bantonsson:
# What does this PR do?

This PR currently hides all the OTel specific things in the sampling code behind traits, so the sampling logic can be moved to a separate crate.

# Motivation

Make the sampling code more generic and move it into a separate crate so it can be reused.

# Additional Notes

The benchmarks are on par with the original code. Some improvements and a few minor regressions. There will be additional changes that move the code out.

### DIFF
--- a/datadog-opentelemetry/src/sampling/otel_mappings.rs
+++ b/datadog-opentelemetry/src/sampling/otel_mappings.rs
@@ -81,10 +81,6 @@ impl<'a> OtelSpan<'a> for PreSampledSpan<'a> {
 
 impl SpanProperties for PreSampledSpan<'_> {
     type Attribute = opentelemetry::KeyValue;
-    type AttributesIter<'b>
-        = std::slice::Iter<'b, opentelemetry::KeyValue>
-    where
-        Self: 'b;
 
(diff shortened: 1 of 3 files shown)


datadog-opentelemetry/src/sampling/datadog_sampler.rs
```diff
@@ -447,98 +431,113 @@ impl DatadogSampler {
...
     }
 }
 
-pub(crate) struct DdSamplingResult {
-    pub is_keep: bool,
-    pub trace_root_info: Option<TraceRootSamplingInfo>,
+pub struct TraceRootSamplingInfo {
+    mechanism: SamplingMechanism,
+    rate: f64,
+    rl_effective_rate: Option<i32>,
```
brettlangdon:
```suggestion
    rl_effective_rate: Option<f32>,
```

? don't we need the rate limiter rate to be a float between 0 and 1?

ekump (approved):
Can't speak to the sampler logic, but just a couple non-blocking questions about some API stuff.

PR #171, opened 2026-02-26
docs: improve log/metric API documentation

paullegranddc:
# What does this PR do?

* Document usage and initilaization of the logs and metrics API better.
* Recommend going through the `log` crate for the logging API, as the otel API is not made to be used directly
* Document that the metrics and logs API are behind feature gates


datadog-opentelemetry/Cargo.toml
```diff
@@ -9,6 +9,13 @@ license.workspace = true
 readme.workspace = true
 authors.workspace = true
 
+[package.metadata.docs.rs]
+# document all features
+feature = ["logs", "metrics"]
```
ekump: doesn't this have to be `features`?
```suggestion
features = ["logs", "metrics"]
```

PR #189, opened 2026-03-17
feat(datadog-aws): add AWS SDK tracing for SQS, SNS, and EventBridge

Dogbu-cyber:
# What does this PR do?

Adds Datadog trace context propagation for the AWS SDK for Rust, covering SQS, SNS, and EventBridge through separate service crates. Each service crate depends only on its matching AWS SDK crate, with shared span and attribute logic in `datadog-aws-core`.

Ref: #190
Jira: [IDMPL-570]

## Crate architecture

```
instrumentation/
├── datadog-aws-core/             # shared request span lifecycle, propagation headers, attribute keys
├── datadog-aws-core-test-utils/  # shared integration/unit test helpers, not published
├── datadog-aws-sqs/              # SQS SDK interceptor and SQS context extraction
├── datadog-aws-sns/              # SNS SDK interceptor
├── datadog-aws-eventbridge/      # EventBridge SDK interceptor
└── datadog-aws-lambda/           # existing Lambda crate, updated for publishing/dependency cleanup
```

## Usage

```rust
use datadog_aws_sns::ConfigExt as _;

let config = aws_sdk_sns::config::Builder::from(&sdk_config)
    .datadog_tracing() // install Datadog tracing interceptor
    .build();

let client = aws_sdk_sns::Client::from_conf(config);
```

## Supported behavior

| Service | Context injection | Service-specific span tags | Injection point |
|---------|-------------------|----------------------------|-----------------|
| SQS | `SendMessage`, `SendMessageBatch` | queue tags for send, receive, and delete operations; batch/message counts where available; send message ID; receive span links | `_datadog` message attribute as String JSON |
| SNS | `Publish`, `PublishBatch` | topic/target tags for publish, batch, topic, subscription, and permission operations; batch counts where available | `_datadog` message attribute as Binary JSON |
| EventBridge | `PutEvents` | rule tags for rule/target operations; batch counts for `PutEvents` | top-level `_datadog` field in each event `detail` JSON object |

Each instrumented AWS call creates a client span named `sqs.request`, `sns.request`, or `eventbridge.request`, with common Datadog/OpenTelemetry attributes including `aws.service`, `aws.operation`, `aws.region`, `aws.partition`, `resource.name`, `operation.name`, `span.kind`, HTTP method/URL/status, AWS user agent, and AWS request ID.

## Notable details

- SQS receive requests automatically request the `_datadog` message attribute, expose `datadog_aws_sqs::extract_context`, and add span links for received messages with valid propagated context.
- SQS and SNS respect the AWS 10-message-attribute limit, but replace an existing stale `_datadog` attribute when present.
- SNS uses Binary message attributes for `_datadog` to avoid SNS subscription filter policy interference.
- EventBridge injects only into existing top-level JSON object details, replaces an existing top-level `_datadog`, preserves nested `_datadog` fields, skips missing/invalid/non-object details, and skips details larger than 1 MiB.
- Injection failures never fail the AWS SDK call; they are skipped and logged at debug level.
- Integration tests cover successful injection, span tags, error status handling, SQS receive links, batch message counts, and EventBridge detail rewrite edge cases.

### Lambda: SQS.SendMessage and EventBridge.PutEvents
image

### SQS.SendMessageBatch
image

### SQS.ReceiveMessage
image

[IDMPL-570]: https://datadoghq.atlassian.net/browse/IDMPL-570?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


joeyzhao2018 (approved):
LGTM

Dogbu-cyber:
@codex review

instrumentation/aws/datadog-aws/tests/integration.rs
```diff
@@ -0,0 +1,766 @@
...
+async fn mock_aws(status: u16) -> (String, JoinHandle<()>, Arc<Mutex<Vec<String>>>) {
+    let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
+    let addr: SocketAddr = listener.local_addr().unwrap();
+    let bodies: Arc<Mutex<Vec<String>>> = Arc::new(Mutex::new(Vec::new()));
+    let bodies_server = bodies.clone();
+    let handle = tokio::spawn(async move {
+        loop {
+            let Ok((stream, _)) = listener.accept().await else {
+                break;
+            };
```
ygree:
```suggestion
        while let Ok((stream, _)) = listener.accept().await {
```

instrumentation/aws/datadog-aws/src/interceptor.rs
```diff
@@ -0,0 +1,277 @@
...
+        let service_id = handler.span_service_id();
+        let sdk_service_name = handler.sdk_service_name();
+        let mut tags = base_tags(service_id, sdk_service_name, operation, region, partition);
+        tags.extend(handler.service_tags(operation, context.input(), region, partition));
+
+        // Create the span as a child of the current context. This must happen
+        // before extracting trace headers so the injected context points to
+        // this span, not the caller's span.
+        let parent_cx = Context::current();
+        let tracer = global::tracer(TRACER_NAME);
```
ygree: `global::tracer(TRACER_NAME)` is called on every AWS SDK invocation, but it acquires an `RwLock` read guard and boxes a new tracer each time for a constant name. Consider resolving it once in `new()` and storing it as an `Arc<BoxedTracer>` field.

instrumentation/aws/datadog-aws/src/interceptor.rs
```diff
@@ -0,0 +1,281 @@
...
+        context: &FinalizerInterceptorContextRef<'_>,
+        _runtime_components: &RuntimeComponents,
+        cfg: &mut ConfigBag,
+    ) -> Result<(), BoxError> {
+        let Some(span_ctx) = cfg.load::<SpanContext>() else {
+            return Ok(());
+        };
+        let span = span_ctx.0.span();
+
+        // Re-set response tags to cover cases where read_after_transmit did not run
```
ygree: `read_after_transmit` seems to be redundant

instrumentation/datadog-aws-core/src/interceptor.rs
```diff
@@ -0,0 +1,340 @@
...
+    ) -> Vec<KeyValue>;
+}
+
+/// Generic AWS SDK interceptor that creates a Datadog span and injects trace
+/// context for the service described by the provided [`ServiceHandler`].
+///
+/// Not intended to be used directly — each service crate exposes a named
+/// wrapper type (`SqsInterceptor`, `SnsInterceptor`, `EventBridgeInterceptor`).
+pub struct AwsInterceptor {
+    handler: Box<dyn ServiceHandler>,
```
ygree: Suggestion: Refactor AwsInterceptor from Box<dyn ServiceHandler> to a generic AwsInterceptor<H>.

```suggestion
pub struct AwsInterceptor<H: ServiceHandler> {
    handler: H,
```

instrumentation/datadog-aws-core/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+repository.workspace = true
+authors.workspace = true
+publish = false
+
+[features]
+test-utils = [
```
ygree: Suggestion: move test utils out to a separate internal crate (datadog-aws-test-utils) in order to keep datadog-aws-core strictly production-focused published parts, make tests-only code and dependencies explicit (dev-dependencies), and reduce packaging/feature complexity (the test-utils feature then can be removed). This should also improve separation of concerns, avoid accidental exposure or coupling of test helpers through a runtime crate.

instrumentation/Cargo.toml
```diff
@@ -2,11 +2,17 @@
 # SPDX-License-Identifier: Apache-2.0
 
 [workspace]
-members = ["datadog-aws", "datadog-aws-lambda"]
+members = [
+    "datadog-aws-core",
```
ygree, line 6: This separation looks good!

instrumentation/datadog-aws-eventbridge/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+        detail_map.insert(DATADOG_ATTRIBUTE_KEY.into(), trace_ctx);
+
+        let new_detail = match serde_json::to_string(&detail_map) {
+            Ok(s) => s,
+            Err(_) => continue,
+        };
+
+        // EventBridge entries have a 1 MB detail size limit.
+        if new_detail.len() > ONE_MB {
+            continue;
```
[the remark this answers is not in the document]
ygree: TL;DR This is a valid concern.  I'm thinking about dropping this check because it's not correct or useful. 

The limit is documented as a service-side contract. AWS’s current EventBridge API reference says `PutEvents` total entry size must be less than 1 MB, and the user guide says the whole request must stay under `1,048,576` bytes, with a single entry allowed to use that full budget if it is the only one. Invalid entries are reported back by the service through `PutEventsResultEntry.ErrorCode` / `ErrorMessage`, which is consistent with backend enforcement rather than SDK preflight. Sources: [PutEvents API](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_PutEvents.html), [PutEvents user guide](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-putevents.html), [PutEventsResultEntry](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_PutEventsResultEntry.html).

[The latest generated Rust SDK docs still say `256KB`](https://docs.rs/aws-sdk-eventbridge/1.105.0/aws_sdk_eventbridge/operation/put_events/builders/struct.PutEventsFluentBuilder.html)

instrumentation/datadog-aws-eventbridge/src/lib.rs
```diff
@@ -0,0 +1,393 @@
...
+                DATADOG_RESOURCE_NAME_KEY.into(),
+                serde_json::Value::String(name.into()),
+            );
+        }
+
+        let trace_ctx = serde_json::Value::Object(entry_ctx);
+
+        let detail = entry.detail.as_deref().unwrap_or("{}");
+        let mut detail_map: serde_json::Map<String, serde_json::Value> =
+            match serde_json::from_str(detail) {
```
ygree: This requires full detail deserialization to an expensive memory representation. There must be a lighter-weight way to achieve adding the `_datadog` data field.

ygree: Fixed in 9f56055e6aeb2d3cd4b9b8a9545ea1de43681680

instrumentation/datadog-aws-eventbridge/src/lib.rs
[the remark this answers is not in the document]
ygree: I think we should keep the check for now, just to provide an upper limit. If the payload exceeds the limit for any reason, we can skip the injection path altogether, as such a request will be rejected by the backend anyway.

instrumentation/datadog-aws-eventbridge/src/lib.rs
```diff
@@ -279,16 +286,13 @@ fn inject_into_put_events(
         let trace_ctx = serde_json::Value::Object(entry_ctx);
 
         let detail = entry.detail.as_deref().unwrap_or("{}");
-        let mut detail_map: serde_json::Map<String, serde_json::Value> =
```
ygree: It used to deserialize entire `detail`. Now it uses a visitor to scan only top level and replace/inject missing context recreating a new detail content.

instrumentation/datadog-aws-eventbridge/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+
+        let detail = entry.detail.as_deref().unwrap_or("{}");
+        if detail.len() > MAX_EVENT_DETAIL_BYTES {
+            continue;
+        }
+
+        let new_detail = match rewrite_json_object_field(detail, DATADOG_ATTRIBUTE_KEY, &trace_ctx)
+        {
+            Ok(detail) => detail,
+            Err(_) => continue,
```
purple4reina: Would be nice to include a debug level log here, or somehow otherwise inform the user why injection failed.

instrumentation/datadog-aws-sns/src/lib.rs
```diff
@@ -0,0 +1,424 @@
...
+        let Ok(datadog_attr) = build_datadog_attribute(trace_headers) else {
+            return;
+        };
+        inject_message_attribute(&mut publish_input.message_attributes, datadog_attr);
+    } else if let Some(batch_input) = input.downcast_mut::<PublishBatchInput>() {
+        let Some(entries) = batch_input.publish_batch_request_entries.as_mut() else {
+            return;
+        };
+        let Ok(dd_attr) = build_datadog_attribute(trace_headers) else {
+            return;
```
purple4reina: In all the cases where we return early, would be nice to have some debug level logging.

purple4reina (approved):

ygree:
@codex review

instrumentation/datadog-aws-sqs/src/lib.rs
```diff
@@ -0,0 +1,604 @@
...
+    message: &Message,
+    propagator: &dyn TextMapPropagator,
+) -> Option<Context> {
+    let trace_headers = datadog_trace_headers(message)?;
+    extract_context_from_trace_headers(&trace_headers, propagator)
+}
+
+fn datadog_trace_headers(message: &Message) -> Option<HashMap<String, String>> {
+    let attrs = message.message_attributes.as_ref()?;
+    let datadog_attr = attrs.get(DATADOG_ATTRIBUTE_KEY)?;
```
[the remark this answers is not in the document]
ygree: Fixed in 1083b6b7b508130796001843784702174bf376f4

instrumentation/datadog-aws-sns/src/lib.rs
```diff
@@ -0,0 +1,447 @@
...
+            topic_arn = input.topic_arn.as_deref();
+        } else if let Some(input) = input.downcast_ref::<ListSubscriptionsByTopicInput>() {
+            topic_arn = input.topic_arn.as_deref();
+        } else if let Some(input) = input.downcast_ref::<RemovePermissionInput>() {
+            topic_arn = input.topic_arn.as_deref();
+        } else if let Some(input) = input.downcast_ref::<SetTopicAttributesInput>() {
+            topic_arn = input.topic_arn.as_deref();
+        } else if let Some(input) = input.downcast_ref::<SubscribeInput>() {
+            topic_arn = input.topic_arn.as_deref();
+        }
```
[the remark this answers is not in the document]
ygree: Fixed in b35e9ffd12c25b308660738d87c92db26b9761c8

instrumentation/datadog-aws-sqs/src/lib.rs
```diff
@@ -0,0 +1,604 @@
...
+        } else if let Some(input) = input.downcast_ref::<SendMessageBatchInput>() {
+            queue_url = input.queue_url.as_deref();
+            batch_message_count = Some(input.entries.as_ref().map_or(0, Vec::len) as i64);
+        } else if let Some(input) = input.downcast_ref::<ReceiveMessageInput>() {
+            queue_url = input.queue_url.as_deref();
+        } else if let Some(input) = input.downcast_ref::<DeleteMessageInput>() {
+            queue_url = input.queue_url.as_deref();
+        } else if let Some(input) = input.downcast_ref::<DeleteMessageBatchInput>() {
+            queue_url = input.queue_url.as_deref();
+            batch_message_count = Some(input.entries.as_ref().map_or(0, Vec::len) as i64);
```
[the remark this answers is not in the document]
ygree: Fixed in b35e9ffd12c25b308660738d87c92db26b9761c8

bantonsson:
Hey @ygree is can this PR be moved to _draft_ or does it need a full review from @DataDog/apm-rust?

instrumentation/datadog-aws-eventbridge/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+repository.workspace = true
+authors.workspace = true
+publish.workspace = true
+
+[dependencies]
+datadog-aws-core = { workspace = true }
+aws-sdk-eventbridge = "1.104.0"
```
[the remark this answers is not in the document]
ygree: Addressing in a separate PR #297

instrumentation/datadog-aws-sns/src/lib.rs
```diff
@@ -0,0 +1,447 @@
...
+    }
+}
+
+fn inject_message_attribute(
+    message_attributes: &mut Option<HashMap<String, MessageAttributeValue>>,
+    datadog_attr: MessageAttributeValue,
+) {
+    let attrs = message_attributes.get_or_insert_with(HashMap::new);
+    if attrs.len() < MAX_MESSAGE_ATTRIBUTES || attrs.contains_key(DATADOG_ATTRIBUTE_KEY) {
+        attrs.insert(DATADOG_ATTRIBUTE_KEY.to_string(), datadog_attr);
```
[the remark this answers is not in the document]
ygree: Exact message-size validation is intentionally left to AWS. The required injection headroom is already documented in d6eb725a13649d2835d0f8814c4b087774b193f0, and the existing tracer behavior is inconsistent rather than a clear precedent for duplicating backend quota logic: JS has a proper EventBridge full request-size guard, Python has a stale/partial EventBridge guard, Java has no EventBridge size guard found, and SQS/SNS across JS, Python, and Java appear to use only attribute-count guards. Because a size gate would only affect payloads already extremely close to the service limit while adding AWS-specific accounting that can drift over time, we should keep the stable count/validity guards and let AWS enforce exact serialized size limits.

instrumentation/datadog-aws-sqs/src/lib.rs
```diff
@@ -0,0 +1,604 @@
...
+    }
+}
+
+fn inject_message_attribute(
+    message_attributes: &mut Option<HashMap<String, MessageAttributeValue>>,
+    datadog_attr: MessageAttributeValue,
+) {
+    let attrs = message_attributes.get_or_insert_with(HashMap::new);
+    if attrs.len() < MAX_MESSAGE_ATTRIBUTES || attrs.contains_key(DATADOG_ATTRIBUTE_KEY) {
+        attrs.insert(DATADOG_ATTRIBUTE_KEY.to_string(), datadog_attr);
```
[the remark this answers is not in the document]
ygree: Exact message-size validation is intentionally left to AWS. The required injection headroom is already documented in d6eb725a13649d2835d0f8814c4b087774b193f0, and the existing tracer behavior is inconsistent rather than a clear precedent for duplicating backend quota logic: JS has a proper EventBridge full request-size guard, Python has a stale/partial EventBridge guard, Java has no EventBridge size guard found, and SQS/SNS across JS, Python, and Java appear to use only attribute-count guards. Because a size gate would only affect payloads already extremely close to the service limit while adding AWS-specific accounting that can drift over time, we should keep the stable count/validity guards and let AWS enforce exact serialized size limits.

ygree:
> Hey @ygree is can this PR be moved to _draft_ or does it need a full review from @DataDog/apm-rust?

Moving it back from _draft_ as it's ready for a full review from @DataDog/apm-rust now

instrumentation/datadog-aws-sqs/src/lib.rs
```diff
@@ -0,0 +1,1194 @@
...
+/// Use [`ConfigExt::datadog_tracing`] to install it on an SQS config builder.
+#[derive(Debug)]
+struct SqsInterceptor {
+    tracer: global::BoxedTracer,
+}
+
+impl SqsInterceptor {
+    fn new() -> Self {
+        Self {
+            tracer: global::tracer(TRACER_NAME),
```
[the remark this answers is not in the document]
ygree: Fixed in 9087bb8c9b180e9aa209cc964837171b8f02f002.

instrumentation/Cargo.toml
```diff
@@ -2,23 +2,33 @@
...
 
 [workspace.package]
-version = "0.1.0"
 edition = "2021"
 rust-version = "1.91.1"
 license = "Apache-2.0"
 repository = "https://github.com/DataDog/dd-trace-rs"
 authors = ["Datadog Inc. <info@datadoghq.com>"]
-publish = false
+publish = true
```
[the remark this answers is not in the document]
ygree: Fixed in 965e3e8d190c373ecbaf63972ff0b9eba73cae8 by turning the publish off for the instrumentation crates for now.

instrumentation/datadog-aws-eventbridge/src/lib.rs
```diff
@@ -0,0 +1,573 @@
...
+            rule_name = input.name.as_deref();
+        } else if let Some(input) = input.downcast_ref::<EnableRuleInput>() {
+            rule_name = input.name.as_deref();
+        } else if let Some(input) = input.downcast_ref::<DisableRuleInput>() {
+            rule_name = input.name.as_deref();
+        } else if let Some(input) = input.downcast_ref::<PutTargetsInput>() {
+            rule_name = input.rule.as_deref();
+        } else if let Some(input) = input.downcast_ref::<RemoveTargetsInput>() {
+            rule_name = input.rule.as_deref();
+        } else if let Some(input) = input.downcast_ref::<PutEventsInput>() {
```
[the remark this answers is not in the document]
ygree: Fixed in df2ba0ba64296622ab75ea3a29e387d342b378cd.

instrumentation/datadog-aws-sqs/src/lib.rs
```diff
@@ -0,0 +1,1194 @@
...
+        cfg: &mut ConfigBag,
+    ) -> Result<(), BoxError> {
+        let Some(metadata) = AwsRequestMetadata::from_config_bag(cfg) else {
+            return Ok(());
+        };
+
+        let input = context.input();
+        let queue_url = queue_url_from_input(input);
+        let batch_message_count = batch_message_count_from_input(input);
+        let mut queue_name = None;
```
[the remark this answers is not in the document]
ygree: Fixed in ef7477fd3ee83cc3e26226a3ab91d4bf169ce8ad.

instrumentation/datadog-aws-core/src/attribute_keys.rs
```diff
@@ -0,0 +1,38 @@
+// Copyright 2025-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+// Trace propagation payload keys.
+pub const DATADOG_ATTRIBUTE_KEY: &str = "_datadog";
```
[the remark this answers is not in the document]
ygree: Fixed in a10f561e538c9da7e035be7b0f7e9779fa5112e4.

instrumentation/datadog-aws-core/src/request_span.rs
```diff
@@ -0,0 +1,187 @@
...
+    } else if region.starts_with("us-isof-") {
+        "aws-iso-f"
+    } else if region.starts_with("us-isob-") {
+        "aws-iso-b"
+    } else if region.starts_with("us-iso-") {
+        "aws-iso"
+    } else if region.starts_with("eu-isoe-") {
+        "aws-iso-e"
+    } else {
+        "aws"
```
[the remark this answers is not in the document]
ygree: Fixed in 5db9dd20fffe85dd9f2fd04f53360cb2823eb206.

ygree:
@codex review

instrumentation/datadog-aws-sqs/src/lib.rs
```diff
@@ -0,0 +1,1221 @@
...
+    } else {
+        None
+    }
+}
+
+fn batch_message_count_from_input(input: &Input) -> Option<i64> {
+    if let Some(input) = input.downcast_ref::<SendMessageBatchInput>() {
+        return Some(input.entries.as_ref().map_or(0, Vec::len) as i64);
+    }
+
```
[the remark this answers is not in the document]
ygree: Fixed in ec31b0e473498b0269a4bedc640e7853b2f9b12d

instrumentation/datadog-aws-sqs/src/lib.rs
```diff
@@ -0,0 +1,1221 @@
...
+        input.queue_url.as_deref()
+    } else if let Some(input) = input.downcast_ref::<SetQueueAttributesInput>() {
+        input.queue_url.as_deref()
+    } else if let Some(input) = input.downcast_ref::<TagQueueInput>() {
+        input.queue_url.as_deref()
+    } else if let Some(input) = input.downcast_ref::<UntagQueueInput>() {
+        input.queue_url.as_deref()
+    } else {
+        None
+    }
```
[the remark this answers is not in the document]
ygree: Extracting queue_url from ARN is probably excessive. It could be implemented in the future, but there is currently no useful scenario for it. Other instrumentations, such as dd-trace-java don't implement this either. Additionally, CancelMessageMoveTask doesn't event have an ARN.

instrumentation/datadog-aws-core-test-utils/src/lib.rs
```diff
@@ -0,0 +1,5 @@
+// Copyright 2025-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+pub mod integration_test_helpers;
+pub mod test_helpers;
```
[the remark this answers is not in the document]
ygree: Fixed in adc7115a5f3fe9966d88eb53db75a3b8894f1cf5

ygree:
@codex review

instrumentation/datadog-aws-sqs/src/lib.rs
```diff
@@ -0,0 +1,1271 @@
...
+    extract_context_from_trace_headers(&trace_headers, propagator)
+}
+
+fn datadog_trace_headers(message: &Message) -> Option<HashMap<String, String>> {
+    if let Some(datadog_attr) = message
+        .message_attributes
+        .as_ref()
+        .and_then(|attrs| attrs.get(DATADOG_ATTRIBUTE_KEY))
+    {
+        return trace_headers_from_sqs_message_attribute(datadog_attr);
```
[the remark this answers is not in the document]
ygree: Fixed in fe9a88f2ccdb642a57de4861f0278709deb96fcd

ygree:
@codex review

instrumentation/datadog-aws-core-test-utils/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+version = "0.1.0"
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+repository.workspace = true
+authors.workspace = true
+publish = false
+
+[dependencies]
+aws-credential-types = { version = "1", features = ["test-util"] }
```
paullegranddc: why is this one not a workspace dependency like the rest?

instrumentation/datadog-aws-eventbridge/src/lib.rs
```diff
@@ -0,0 +1,569 @@
...
+//! # Payload Size Headroom
+//!
+//! Datadog trace context is injected into AWS payload fields before the SDK sends the request, so
+//! applications should leave a small amount of room under the EventBridge `PutEvents` size limit.
+//! With the default W3C trace-context propagator, the inserted JSON is typically less than 100
+//! bytes before AWS request serialization overhead. `event_bus_name` adds the encoded bus name plus
+//! JSON field overhead, and baggage from the configured global OpenTelemetry text-map propagator
+//! can add arbitrary bytes. EventBridge performs the authoritative size validation, and this crate
+//! only applies cheap stable guards such as skipping invalid or clearly oversized `detail`
+//! payloads.
```
paullegranddc: As a user this is interesting to know, but do we want this to be at the top level documentation when user encouter the crate?

instrumentation/datadog-aws-eventbridge/src/lib.rs
```diff
@@ -0,0 +1,569 @@
...
+//! # Usage
+//!
+//! ```rust,ignore
+//! use datadog_aws_eventbridge::ConfigExt as _;
+//!
+//! let config = aws_sdk_eventbridge::config::Builder::from(&sdk_config)
+//!     .datadog_tracing()
+//!     .build();
+//! let client = aws_sdk_eventbridge::Client::from_conf(config);
+//! ```
```
paullegranddc, line 37: Same remark for every crate, but I think Usage should be the 2nd section just after `What it instruments`

instrumentation/datadog-aws-eventbridge/src/lib.rs
```diff
@@ -0,0 +1,569 @@
...
+//! With the default W3C trace-context propagator, the inserted JSON is typically less than 100
+//! bytes before AWS request serialization overhead. `event_bus_name` adds the encoded bus name plus
+//! JSON field overhead, and baggage from the configured global OpenTelemetry text-map propagator
+//! can add arbitrary bytes. EventBridge performs the authoritative size validation, and this crate
+//! only applies cheap stable guards such as skipping invalid or clearly oversized `detail`
+//! payloads.
+//!
+//! # Usage
+//!
+//! ```rust,ignore
```
paullegranddc: Prefer ```rust,no_run instead of ```run,ignore everywhere in the crates public dox for two reasons:
* The compiler still checks that it compiles, even if it doesn't run it as tests
* It removes this yellow warning next to the snippet

Image

instrumentation/datadog-aws-sqs/src/lib.rs
```diff
@@ -0,0 +1,1354 @@
...
+//! `cloud.resource_id`, `messaging.batch.message_count`, and the sent `messaging.message.id` when
+//! available.
+//!
+//! For `SendMessage` and `SendMessageBatch`, the interceptor injects the active request span
+//! context as an SQS message attribute named `_datadog` with data type `String`. Existing
+//! `_datadog` attributes are replaced. Injection is skipped when the message already has the
+//! maximum number of SQS message attributes and no `_datadog` attribute is present.
+//!
+//! For `ReceiveMessage`, the interceptor requests the `_datadog` message attribute if the caller
+//! did not already request it. After messages are received, the request span records the received
```
paullegranddc: > if the caller did not already request it

Maybe it's because I am not familiar with the sqs API, but I don't understand what that means

instrumentation/datadog-aws-eventbridge/Cargo.toml
paullegranddc: same here, why is this not a workspace dep as the rest?

paullegranddc (commented):
Also if we want to push these to crates.io, we need to put README at the crate root as this is going to be used for the crate page.
But this can be done in a followup PR

instrumentation/datadog-aws-eventbridge/src/lib.rs
ygree: Good to know! Thanks!

instrumentation/datadog-aws-eventbridge/src/lib.rs
ygree: I'll keep the short limitations paragraph under the usage block and move this longer explanation to the datadog_tracing function docs.

instrumentation/datadog-aws-eventbridge/src/lib.rs
ygree: Agreed!

instrumentation/datadog-aws-core-test-utils/Cargo.toml
ygree: The main reason is that such dependencies have only been used once in a single crate instead of being shared across multiple crates so far. 

At the same time, centralizing them can make updates easier and allow for easier review in one place. So, I'll move the version declaration of such dependencies to the workspace level.

instrumentation/datadog-aws-sqs/src/lib.rs
ygree: Good catch! I've improved it.

paullegranddc (approved):

ygree:
> Also if we want to push these to crates.io, we need to put README at the crate root as this is going to be used for the crate page. But this can be done in a followup PR

Done!

PR #200, opened 2026-03-27
feat(baggage)!: add support for w3c baggage

rachelyangdog:
# What does this PR do?

Adds W3C Baggage propagation support (baggage header) to the composite propagator. When baggage is included in the propagation style configuration, the global TextMapPropagator will read and write the baggage HTTP header using opentelemetry_sdk::propagation::BaggagePropagator.
                                                                                                                                                                                                                                              
Changes:                                                                                                                                                                                                                                  
-  Adds TracePropagationStyle::Baggage variant, recognized case-insensitively from `DD_TRACE_PROPAGATION_STYLE`, `DD_TRACE_PROPAGATION_STYLE_EXTRACT`, and `DD_TRACE_PROPAGATION_STYLE_INJECT`
- Adds baggage to the default propagation styles alongside datadog and tracecontext                                                                                                                                                         
- Wires inject/extract separately at the DatadogPropagator layer, respecting the extract-vs-inject style override independently
- Adds propagation::baggage module exposing the header key so fields() includes "baggage" when the style is configured

- [system-tests PR ](https://github.com/DataDog/system-tests/pull/6686)

# Motivation

# Additional Notes
- No new dependencies: opentelemetry_sdk::propagation::BaggagePropagator was already a transitive dependency                                                                                                                               
- Baggage propagation is orthogonal to trace context: it operates on OTel Context directly rather than through the internal Propagator<C> trait (which is trace-ID-centric). The Baggage variant's extract/inject arms in that trait are intentional no-ops; the actual work happens one layer up in DatadogPropagator.                                                                                                                                                                
- `DD_TRACE_PROPAGATION_STYLE_EXTRACT` and `DD_TRACE_PROPAGATION_STYLE_INJECT` can be used to enable baggage on only one direction independently of the other

### DIFF
--- a/datadog-opentelemetry/examples/propagator/src/server.rs
+++ b/datadog-opentelemetry/examples/propagator/src/server.rs
@@ -121,7 +121,7 @@ async fn send_request(
 
     let cx = Context::current().with_baggage(vec![
         KeyValue::new("request-id", "xyz-123"),
-        KeyValue::new("service", "rust-propagator-service-example"),
+        KeyValue::new("caller", "rust-propagator-example"),
     ]);
 
     let mut req = hyper::Request::builder().uri(url);
--- a/datadog-opentelemetry/examples/propagator/src/server.rs
+++ b/datadog-opentelemetry/examples/propagator/src/server.rs
@@ -119,7 +119,10 @@ async fn send_request(
 ) -> std::result::Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
     let client = Client::builder(TokioExecutor::new()).build_http();
 
-    let cx = Context::current();
+    let cx = Context::current().with_baggage(vec![
+        KeyValue::new("request-id", "xyz-123"),
+        KeyValue::new("service", "rust-propagator-service-example"),
+    ]);
 
     let mut req = hyper::Request::builder().uri(url);
     global::get_text_map_propagator(|propagator| {
--- a/datadog-opentelemetry/src/core/configuration/configuration.rs
+++ b/datadog-opentelemetry/src/core/configuration/configuration.rs
@@ -773,6 +773,8 @@ pub enum TracePropagationStyle {
     Datadog,
     /// W3C Trace Context propagation format using `traceparent` and `tracestate` headers.
     TraceContext,
+    /// W3C Baggage propagation format using the `baggage` header.
+    Baggage,
     /// No propagation - trace context is not propagated.
     None,
 }
@@ -804,6 +806,7 @@ impl FromStr for TracePropagationStyle {
         match s.trim().to_lowercase().as_str() {
             "datadog" => Ok(TracePropagationStyle::Datadog),
             "tracecontext" => Ok(TracePropagationStyle::TraceContext),
+            "baggage" => Ok(TracePropagationStyle::Baggage),
             "none" => Ok(TracePropagationStyle::None),
             _ => Err(format!("Unknown trace propagation style: '{s}'")),
         }
@@ -815,6 +818,7 @@ impl Display for TracePropagationStyle {
         let style = match self {
             TracePropagationStyle::Datadog => "datadog",
             TracePropagationStyle::TraceContext => "tracecontext",
+            TracePropagationStyle::Baggage => "baggage",
             TracePropagationStyle::None => "none",
         };
         write!(f, "{style}")
@@ -1822,6 +1826,7 @@ fn default_config() -> Config {
             Some(vec![
                 TracePropagationStyle::Datadog,
                 TracePropagationStyle::TraceContext,
+                TracePropagationStyle::Baggage,
             ]),
         ),
         trace_propagation_style_extract: ConfigItem::new(
@@ -2726,6 +2731,7 @@ mod tests {
             Some(vec![
                 TracePropagationStyle::Datadog,
                 TracePropagationStyle::TraceContext,
+                TracePropagationStyle::Baggage,
             ])
             .as_deref()
         );
@@ -2735,6 +2741,47 @@ mod tests {
             Some(vec![TracePropagationStyle::TraceContext]).as_deref()
         );
         assert!(config.trace_propagation_extract_first());
+    }
+
+    #[test]
+    fn test_propagation_style_baggage_parsed_from_env() {
+        // "baggage" is recognised as a valid style value (case-insensitive) in all three env vars.
+        let mut sources = CompositeSource::new();
+        sources.add_source(HashMapSource::from_iter(
+            [
+                ("DD_TRACE_PROPAGATION_STYLE", "datadog,tracecontext,baggage"),
+                ("DD_TRACE_PROPAGATION_STYLE_EXTRACT", "Baggage,datadog"),
+                ("DD_TRACE_PROPAGATION_STYLE_INJECT", "BAGGAGE,tracecontext"),
+            ],
+            ConfigSourceOrigin::EnvVar,
+        ));
+        let config = Config::builder_with_sources(&sources).build();
+
+        assert_eq!(
+            config.trace_propagation_style(),
+            Some(vec![
+                TracePropagationStyle::Datadog,
+                TracePropagationStyle::TraceContext,
+                TracePropagationStyle::Baggage,
+            ])
+            .as_deref()
+        );
+        assert_eq!(
+            config.trace_propagation_style_extract(),
+            Some(vec![
+                TracePropagationStyle::Baggage,
+                TracePropagationStyle::Datadog,
+            ])
+            .as_deref()
+        );
+        assert_eq!(
+            config.trace_propagation_style_inject(),
+            Some(vec![
+                TracePropagationStyle::Baggage,
+                TracePropagationStyle::TraceContext,
+            ])
+            .as_deref()
+        );
     }
 
     #[test]
--- a/datadog-opentelemetry/src/text_map_propagator.rs
+++ b/datadog-opentelemetry/src/text_map_propagator.rs
@@ -65,8 +65,10 @@ pub struct DatadogPropagator {
     inner: DatadogCompositePropagator<Config>,
     registry: TraceRegistry,
     cfg: Arc<Config>,
+    baggage_propagator: BaggagePropagator,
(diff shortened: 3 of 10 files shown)


datadog-opentelemetry/src/text_map_propagator.rs
```diff
@@ -679,4 +706,272 @@ pub mod tests {
             }
         }
     }
+
+    const BAGGAGE_KEY: &str = "baggage";
```
bantonsson: Isn't this a duplicate definition of https://github.com/DataDog/dd-trace-rs/pull/200/changes#diff-e1b43d6d50252f9d55a51567ae090d317e315132079498afd34c06c41688c592R14

datadog-opentelemetry/examples/propagator/src/server.rs
```diff
@@ -119,7 +119,10 @@ async fn send_request(
 ) -> std::result::Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
     let client = Client::builder(TokioExecutor::new()).build_http();
 
-    let cx = Context::current();
+    let cx = Context::current().with_baggage(vec![
+        KeyValue::new("request-id", "xyz-123"),
+        KeyValue::new("service", "rust-propagator-service-example"),
```
bantonsson: I think it's a bit confusing to have a baggage item with a key `service` that overlaps with a known Datadog span tag, since they are completely separate.

datadog-opentelemetry/src/text_map_propagator.rs
```diff
@@ -83,6 +98,10 @@ impl DatadogPropagator {
         cx: &opentelemetry::Context,
         mut injector: &mut dyn opentelemetry::propagation::Injector,
     ) {
+        if self.baggage_inject {
+            BaggagePropagator::new().inject_context(cx, injector);
```
bantonsson: Can't we reuse the `BaggagePropagator`?

datadog-opentelemetry/src/text_map_propagator.rs
```diff
@@ -167,7 +187,13 @@ impl DatadogPropagator {
                 cx.with_remote_span_context(otel_span_context)
                     .with_value(DatadogExtractData::from_span_context(dd_span_context))
             })
-            .unwrap_or_else(|| cx.clone())
+            .unwrap_or_else(|| cx.clone());
+
+        if self.baggage_extract {
+            BaggagePropagator::new().extract_with_context(&cx, extractor)
```
bantonsson: Same here. Reuse instead of `BaggagePropagator::new()`?

datadog-opentelemetry/src/text_map_propagator.rs
```diff
@@ -679,4 +706,272 @@ pub mod tests {
...
+                .set_trace_propagation_style(styles)
+                .build(),
+        );
+        DatadogPropagator::new(config.clone(), TraceRegistry::new(config))
+    }
+
+    // ── Unit: composite propagator inclusion/exclusion ────────────────────────
+
+    #[test]
+    fn baggage_included_in_fields_when_configured() {
```
bantonsson, line 814: What happens when we only have baggage as an injector or extractor?

bantonsson (commented):
Looks good overall

datadog-opentelemetry/src/text_map_propagator.rs
```diff
@@ -64,14 +65,41 @@ pub struct DatadogPropagator {
...
+    fields: Vec<String>,
 }
 
 impl DatadogPropagator {
     pub(crate) fn new(config: Arc<Config>, registry: TraceRegistry) -> Self {
+        let baggage_extract = config
+            .trace_propagation_style_extract()
+            .or_else(|| config.trace_propagation_style())
+            .unwrap_or_default()
+            .contains(&TracePropagationStyle::Baggage);
```
iunanua: Changing the visibility of [get_extractors](https://github.com/DataDog/dd-trace-rs/blob/main/datadog-opentelemetry/src/propagation/config.rs#L8) fn to `pub(crate)` you could use it here 
```suggestion
        let baggage_extract = get_extractors(config.as_ref()).contains(&TracePropagationStyle::Baggage);
```

datadog-opentelemetry/src/text_map_propagator.rs
```diff
@@ -64,14 +65,41 @@ pub struct DatadogPropagator {
...
+        let baggage_inject = config
+            .trace_propagation_style_inject()
+            .or_else(|| config.trace_propagation_style())
+            .unwrap_or_default()
+            .contains(&TracePropagationStyle::Baggage);
```
iunanua: Same here but changing `get_injectors` visibility
```suggestion
        let baggage_inject = get_injectors(config.as_ref()).contains(&TracePropagationStyle::Baggage);
```

datadog-opentelemetry/src/text_map_propagator.rs
```diff
@@ -64,14 +65,41 @@ pub struct DatadogPropagator {
...
+        let inner = DatadogCompositePropagator::new(config.clone());
+        let mut fields = inner.keys().to_vec();
+        if baggage_inject
```
iunanua: should we take `baggage_inject` or `baggage_extract` into account?
I'm asking because [here](https://github.com/DataDog/dd-trace-rs/blob/main/datadog-opentelemetry/src/propagation/mod.rs#L102) the extractors are the ones setting the keys...

or it doesn't matter...

datadog-opentelemetry/src/text_map_propagator.rs
```diff
@@ -83,6 +111,10 @@ impl DatadogPropagator {
         cx: &opentelemetry::Context,
         mut injector: &mut dyn opentelemetry::propagation::Injector,
     ) {
+        if self.baggage_inject {
+            self.baggage_propagator.inject_context(cx, injector);
```
iunanua, line 102: This is before the `if !otel_span_context.is_valid()`  check (5 lines below).
 
Should we inject the baggage in a invalid span_context?

datadog-opentelemetry/src/text_map_propagator.rs
rachelyangdog: `fields()` per the OTel spec represents the fields this propagator writes to (injection only), so baggage_inject is the right flag here. I had a test asserting that fields() includes baggage when it's only in extract styles, but that was incorrect. I removed it. Thanks for calling this out!

datadog-opentelemetry/src/text_map_propagator.rs
rachelyangdog: This is intentional because baggage propagation is independent of trace context per the W3C spec. A baggage header can and should be injected even without a valid span context (e.g. for propagating application-level metadata like user ID or tenant when tracing isn't active). The is_valid() guard only gates the Datadog/TraceContext headers below it. This test case is covered by `baggage_only_style_injects_only_baggage_header`.

iunanua (approved):
LGTM

datadog-opentelemetry/src/text_map_propagator.rs
```diff
@@ -64,14 +66,39 @@ pub struct DatadogPropagator {
...
+        let baggage_extract =
+            get_extractors(config.as_ref()).contains(&TracePropagationStyle::Baggage);
+        let baggage_inject =
+            get_injectors(config.as_ref()).contains(&TracePropagationStyle::Baggage);
+        let inner = DatadogCompositePropagator::new(config.clone());
+        // Build fields() from injectors only (per OTel spec, fields() represents what the
+        // propagator *writes*). inner.keys() is extractor-derived, so we strip the baggage key
+        // from it and re-add it solely based on baggage_inject.
+        let mut fields: Vec<String> = inner
+            .keys()
```
bantonsson: Why is this not using `inner.fields()`? If there is something wrong with the inner fields then that should be fixed.

datadog-opentelemetry/src/core/configuration/configuration.rs
```diff
@@ -1822,6 +1826,7 @@ fn default_config() -> Config {
             Some(vec![
                 TracePropagationStyle::Datadog,
                 TracePropagationStyle::TraceContext,
+                TracePropagationStyle::Baggage,
```
bantonsson, line 1829: Since this is a change in behavior, it should be flagged as a breaking change `feat(baggage)!:...`

bantonsson (changes requested):

datadog-opentelemetry/src/text_map_propagator.rs
iunanua: By [this](https://github.com/DataDog/dd-trace-rs/pull/200#discussion_r3059295943) comment, I have realized we are obtaining inner fields from extractors and as @rachelyangdog says they correspond with the [injected headers](https://opentelemetry.io/docs/specs/otel/context/api-propagators/#fields-1)

iunanua: It's something we inherited from [bottlecap](https://github.com/DataDog/datadog-lambda-extension/blob/v76/bottlecap/src/traces/propagation/carrier.rs#L16) :thinking:

iunanua: Because is the Otel Extrator trait which defines that [keys() method](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/propagation/mod.rs#L46) 
:confused:

iunanua: Ok, I think get it. They are different things
We have to change the inner fields to fit the injectors.

iunanua: Fix [here](https://github.com/DataDog/dd-trace-rs/pull/217)
Thanks to both of you for bringing this issue to light!

datadog-opentelemetry/src/core/configuration/configuration.rs
paullegranddc: We should have made this enum `#[non_exhaustive]` 🫤

paullegranddc (approved):

bantonsson (approved):

PR #203, opened 2026-04-01
feat(propagation): add `_unstable_propagation` feature to expose propagation module

duncanista:
## Summary

- Adds a `_unstable_propagation` feature flag that makes the `propagation` module publicly accessible without enabling `test-utils` and its heavy dependency chain (criterion, metrics, logs transports, etc.)
- The module is `#[doc(hidden)]` when exposed via `_unstable_propagation`, keeping it out of public documentation
- Intended for internal downstream consumers like `datadog-lambda-extension` that need direct access to propagation types (carrier, context, datadog/W3C headers, `DatadogCompositePropagator`, `PropagationConfig`)

### Usage

```toml
datadog-opentelemetry = { version = "...", features = ["_unstable_propagation"] }
```

## Test plan

- [x] `cargo check` with default features
- [x] `cargo check --features _unstable_propagation`
- [x] `cargo check --features test-utils`
- [x] `cargo clippy -- -D warnings` for default, `_unstable_propagation`, and `_unstable_propagation,test-utils` combinations


datadog-opentelemetry/Cargo.toml
```diff
@@ -106,6 +106,7 @@ logs-http = [
     "opentelemetry-otlp/http-proto",
     "opentelemetry-otlp/reqwest-blocking-client",
 ]
+_propagation = []
```
paullegranddc: I would prefer something a bit scarier like `_unstable_propagation`

paullegranddc (commented):
Other than the naming thing, LGTM

datadog-opentelemetry/Cargo.toml
duncanista: On it!

paullegranddc (approved):

PR #204, opened 2026-04-01
feat(config): unix socket detection

iunanua:
# What does this PR do?

- If `trace_agent_url` is empty and if user hasn't provided `agent_host` nor `agent_port` and UDS is alive, use `/var/run/datadog/apm.socket` as agent_url and `/var/run/datadog/dsd.socket` as  dogstatsd url.
- Change remote config client to support UDS

### Extra
TelemetryWorkerBuilder could be using the wrong host.


datadog-opentelemetry/src/core/telemetry.rs
```diff
@@ -239,8 +239,7 @@ fn make_telemetry_worker(
     custom_handle: Option<Box<dyn TelemetryHandle>>,
 ) -> Result<Box<dyn TelemetryHandle>, Error> {
     if custom_handle.is_none() {
-        let mut builder = worker::TelemetryWorkerBuilder::new(
-            config.trace_agent_url().to_string(),
+        let mut builder = worker::TelemetryWorkerBuilder::new_fetch_host(
```
paullegranddc, line 242: Why this change?

iunanua: Cause it might be wrong since the beginning. We were initializing the builder with `trace_agent_url` as the host. 

But `TelemetryWorkerBuilder::new_fetch_host` infers the host based on data from the real host (hostname, container id, etc) https://github.com/DataDog/libdatadog/blob/main/libdd-telemetry/src/lib.rs#L22

paullegranddc (approved):

PR #210, opened 2026-04-15
feat: publish tracer metadata upon global init

yannham:
# What does this PR do?

This PR hooks up `store_tracer_metadata` from `libdd-library-config` into tracer initialization. Without changing the API or anything else, this will automatically publish tracer metadata as resource attributes both through a pre-existing Datadog-specific protocol, and through the OTel process context.

# Motivation

This is a required step toward supporting ePBF profiling in dd-trace-rs.

# Additional Notes

- I couldn't find attributes for `hostname`, `process_tags` and `container_id` in the config. Happy to be pointed at which value we should use for those (or just let them initialized by default)
- While it's not really expected that people set up multiple tracers in a single app, it's permitted by the dd-trace-rs API. The process context is global to the whole process, hence we can ask when should we set or update this context. The answer of this PR is to do it automatically when a tracer is set globally through `init()`, which should be the common way of doing things, but do not do it for local initialization through `init_local`. This is somehow arbitrary and can be debated.

### DIFF
--- a/LICENSE-3rdparty.csv
+++ b/LICENSE-3rdparty.csv
@@ -54,6 +54,7 @@ digest,https://github.com/RustCrypto/traits,MIT OR Apache-2.0,RustCrypto Develop
 displaydoc,https://github.com/yaahc/displaydoc,MIT OR Apache-2.0,Jane Lusby <jlusby@yaah.dev>
 either,https://github.com/rayon-rs/either,MIT OR Apache-2.0,bluss
 equivalent,https://github.com/indexmap-rs/equivalent,Apache-2.0 OR MIT,The equivalent Authors
+errno,https://github.com/lambda-fairy/rust-errno,MIT OR Apache-2.0,"Chris Wong <lambda.fairy@gmail.com>, Dan Gohman <dev@sunfishcode.online>"
 event-listener,https://github.com/smol-rs/event-listener,Apache-2.0 OR MIT,"Stjepan Glavina <stjepang@gmail.com>, John Nunley <dev@notgull.net>"
 event-listener-strategy,https://github.com/smol-rs/event-listener-strategy,Apache-2.0 OR MIT,John Nunley <dev@notgull.net>
 fnv,https://github.com/servo/rust-fnv,Apache-2.0  OR  MIT,Alex Crichton <alex@alexcrichton.com>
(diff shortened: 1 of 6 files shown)


datadog-opentelemetry/Cargo.toml
```diff
@@ -26,6 +26,7 @@ opentelemetry-otlp = { workspace = true, features = [
     "metrics",
     "logs",
 ], default-features = false, optional = true }
+libdd-library-config = { workspace = true }
```
bantonsson: Is this only used on linux? Can we make it linux only?

datadog-opentelemetry/src/core/configuration/configuration.rs
```diff
@@ -1607,6 +1610,21 @@ impl Config {
...
+    #[cfg(target_os = "linux")]
+    pub(crate) fn to_tracer_metadata(&self) -> TracerMetadata {
+        TracerMetadata {
+            runtime_id: Some(self.runtime_id.to_owned()),
+            tracer_language: "rust".to_owned(),
+            tracer_version: self.tracer_version.to_owned(),
+            service_name: Some(self.service().to_owned()),
+            service_env: self.env().map(str::to_owned),
+            service_version: self.version().map(str::to_owned),
+            // What about hostname, process_tags and container_id ?
```
bantonsson: Should there be a ticket to track this?

bantonsson (approved):
Looks good. Only minor comments.

datadog-opentelemetry/src/core/configuration/configuration.rs
yannham: Actually, I hope to answer the question before merging the PR :sweat_smile: do you know what should go there? Or if I should just omit it for Rust?

bantonsson: https://github.com/DataDog/libdatadog/blob/main/libdd-common/src/entity_id/unix/container_id.rs

datadog-opentelemetry/src/lib.rs
```diff
@@ -344,10 +344,54 @@ impl DatadogTracingBuilder {
...
+            // This stores the current file descriptor corresponding to the published tracer
+            // metadata, if any. If another tracer is set globally later, we swap the new fd in and
+            // close the previous descriptor.
+            //
+            // `RawFd` is defined as `i32` on almost all but exotic platforms, and -1 is never a
+            // valid descriptor (this is the error value returned by e.g. the `open` syscall).
+            static CURRENT_TRACER_METADATA_FD: AtomicI32 = AtomicI32::new(-1);
+            let tracer_metadata = config.to_tracer_metadata();
+
+            match store_tracer_metadata(&tracer_metadata) {
```
ivoanjo: This is an interesting bit... In particular, every other library that "gained" process context capabilities already had the tracer metadata file descriptor.

In our case dd-trace-rs did not have the file descriptor, so it's getting that for the first time. And that... hmmm is a bit of a smell.

By "smell", I mean that for the others, we didn't want to break compatibility so we're running the old thing side-by-side with the new. But for this new thing, we're adding the legacy option as well -- and that seems a bit unnecessary?

So my suggestion would be to maybe tweak the `store_tracer_metadata` API on the libdatadog side to disable the "old" thing; or maybe introduce a `store_otel_tracer_metadata` variant that only publishes the otel variant?

Of course that would mean waiting for the update on the libdatadog side. We could, at least, close the fd immediately here.

_If you ask me_, I think adding unneeded legacy to dd-trace-rs is probably something we'd want to avoid, but it's more of a decision for y'all.

ivoanjo (commented):
This is cool + was way less than I was expecting! :)

I guess other than maybe adding some of the keys that are missing the non-trivial bit is to decide what to do with the legacy file descriptor thing.

ivoanjo:
Ah, quick question -- the way we're wiring it up means it's always on, right? That's exactly what I think we should be doing (and that's most other implementations are doing, at least when they can), just wanted to confirm to mark it on our [library support matrix](https://docs.google.com/document/d/1IwjjVJzEChcFPcnVV2N5Kkjg-4_Q4v4Q3ojpxntbdvY/edit?pli=1&tab=t.s7dfru3a9hnl).

yannham:
> Ah, quick question -- the way we're wiring it up means it's always on, right? That's exactly what I think we should be doing (and that's most other implementations are doing, at least when they can), just wanted to confirm to mark it on our [library support matrix](https://docs.google.com/document/d/1IwjjVJzEChcFPcnVV2N5Kkjg-4_Q4v4Q3ojpxntbdvY/edit?pli=1&tab=t.s7dfru3a9hnl).

Yes, but with a gotcha. There are several ways to initialize a tracer in dd-trace-rs. The "natural" way is to use `init()`, which registers a global tracer, and this is where the context publication is hooked up - in this case, this is indeed always on.

There's also an `init_local()` API, that returns an `SdkTracerProvider`. This one could or could not be globally set later, I believe and I think it could be set through OTel directly, so we might not even be able to intercept the switch (please correct me if I'm wrong, @bantonsson). It's not really clear what to do in this case, so for now we only install the context in the standard, default workflow.

datadog-opentelemetry/src/lib.rs
yannham: Thanks for the context, I wasn't really sure of the legacy way, and who is the _consumer_ of the tracer metadata. I somehow assumed that there could be a legacy reader out there. Personally if we can scrap the whole file descriptor and atomic business I'm very happy!

I _think_ we can actually do it with the current version of libdatadog, because `TracerMetadata` exposes `to_otel_process_ctx()` which is public, so we should be able to create our metadata, convert them to an otel process context and give that to the `ProcessContext` API directly, without going through `store_tracer_metadata`.

bantonsson:
> There's also an init_local() API, that returns an SdkTracerProvider. This one could or could not be globally set later, I believe and I think it could be set through OTel directly, so we might not even be able to intercept the switch.

Yeah, the only way around that would be to wrap the `SdkTracerProvider` and intercept its use I guess, ie publish on first use of the `TracerProvider`.

yannham:
> Yeah, the only way around that would be to wrap the SdkTracerProvider and intercept its use I guess, ie publish on first use of the TracerProvider.

Got it. IMHO, it's a bit too much magic and doesn't look like it's worth the trouble, especially if the assumption that people should do `init()` in the vast majority of cases holds true. We should just document the behavior very clearly (maybe what's missing here is another comment to `init_local()` that also mentions the difference; currently it's only in `init()`). If really needed we can always add a manual way to publish the context for people who use `init_local()` AND need the context.

datadog-opentelemetry/src/lib.rs
ivoanjo: Even better! To me, we could skip the legacy thing for now -- if anyone asks for it later, we can always add it

ivoanjo:
I think supporting only `init()` is fine -- it looks like the intention is that the `init_local()` is the "I want full + fine control, let me drive and I'll make all the decisions" mode so not doing it by default seems to match with the intention of that API --  it's one more thing you'll have to opt into.

LICENSE-3rdparty.csv
```diff
@@ -180,6 +184,7 @@ rmpv,https://github.com/3Hren/msgpack-rust,MIT,Evgeny Safronov <division494@gmai
 rustc-demangle,https://github.com/rust-lang/rustc-demangle,MIT OR Apache-2.0,Alex Crichton <alex@alexcrichton.com>
 rustc_version,https://github.com/djc/rustc-version-rs,MIT OR Apache-2.0,The rustc_version Authors
 rustc_version_runtime,https://github.com/seppo0010/rustc-version-runtime-rs,MIT,Sebastian Waisbrot <seppo0010@gmail.com>
+rustix,https://github.com/bytecodealliance/rustix,Apache-2.0 WITH LLVM-exception OR Apache-2.0 OR MIT,"Dan Gohman <dev@sunfishcode.online>, Jakub Konka <kubkon@jakubkonka.com>"
```
bantonsson, line 187: Is this transitive from `libdd-library-config`?

bantonsson (approved):

LICENSE-3rdparty.csv
yannham: Ah, I suppose, yes. It's been removed in favor of pure `libc` for different reasons in the current `libdatadog/main`, so it should hopefully go away soon, but hasn't been released yet.

datadog-opentelemetry/src/core/configuration/configuration.rs
yannham: Other fields now filled, process_tags tracked in APMSP-2847

PR #194, opened 2026-03-23
chore(instrumentation): set up instrumentation workspace

Dogbu-cyber:
> **PR Stack:** **#194 (workspace setup)** -> #213 (lambda root span) -> #189 (aws-sdk injection) -> #190 (lambda inferred spans)

# What does this PR do?

Establishes the `instrumentation/` Cargo workspace that houses Datadog AWS instrumentation crates. This is the foundation PR -- #213, #189, and #190 add the implementations on top.

## Workspace structure

```
instrumentation/
├── Cargo.toml                  # shared workspace (resolver = "2", MSRV 1.91.1)
├── datadog-aws-core/           # stub -- internal interceptor machinery (added in #189)
├── datadog-aws-sqs/            # stub -- SqsInterceptor (added in #189)
├── datadog-aws-sns/            # stub -- SnsInterceptor (added in #189)
├── datadog-aws-eventbridge/    # stub -- EventBridgeInterceptor (added in #189)
└── datadog-aws-lambda/         # distributed tracing for Rust Lambda functions (added in #213, #190)
```

Issue #175, opened 2026-03-02
[FEATURE]: AWS Lambda Support


joeyzhao2018:
We are working on adding AWS Lambda Support. Please 👍 if your team uses dd-trace-rs for lambdas to help us prioritize this task! 🙇 Thank you.

PR #213, opened 2026-04-21
feat(datadog-aws-lambda): add root invocation span with OTel tracing

Dogbu-cyber:
> **PR Stack:** #194 (workspace setup) -> **#213 (lambda root span)** -> ~~#190~~#219 (lambda inferred spans)

# What does this PR do?

Implements the core Lambda handler wrapper for `datadog-aws-lambda`. Each invocation is automatically instrumented with an `aws.lambda` root span carrying cold_start, request_id, function metadata, and error recording.

This PR intentionally does **not** include trigger extraction or inferred spans. Those are layered on in ~~#190~~#219 with minimal API surface changes.

## Usage

```rust
lambda_runtime::run(TracedService::new(my_handler, Config::default())).await
```

## What's included

- **`TracedService`** - `tower::Service` that wraps lambda handler with OTel tracing
- **`LambdaSpan`** - `aws.lambda` root span with `cold_start`, `request_id`, `function_arn`, `function_version`, `functionname`, `_dd.origin=lambda`
- **`Invocation`** - start/handler_context/finish lifecycle with error recording
- **`Config`** - `service`/`env`/`version` or full `DatadogTracingBuilder` control
- Lambda-appropriate OTel defaults: synchronous writes, no client-side stats

# Motivation

This PR establishes the root invocation tracing that ~~#190~~#219 builds inferred spans on top of.

Ref: #221

Ref: #175


instrumentation/datadog-aws-lambda/src/invocation.rs
```diff
@@ -0,0 +1,265 @@
...
+
+        tracing::debug!(request_id, "creating invocation root span");
+
+        let parent_cx = Context::current();
+
+        let mut builder = tracer.span_builder(TRACER_NAME);
+        builder.span_kind = Some(SpanKind::Server);
+        let attrs = vec![
+            KeyValue::new(attr::OPERATION_NAME, "aws.lambda"),
+            KeyValue::new(attr::LANGUAGE, "rust"),
```
paullegranddc, line 54: does this really need to be tagged? We already include this in multiple tags across the pipeline

instrumentation/datadog-aws-lambda/src/invocation.rs
```diff
@@ -0,0 +1,265 @@
...
+            KeyValue::new(attr::FUNCTION_ARN, lambda_cx.invoked_function_arn.clone()),
+            KeyValue::new(attr::FUNCTION_VERSION, lambda_cx.env_config.version.clone()),
+            KeyValue::new(attr::FUNCTION_NAME, function_name.to_lowercase()),
+            KeyValue::new(attr::RESOURCE_NAMES, function_name.clone()),
+            KeyValue::new(attr::DD_ORIGIN, "lambda"),
+        ];
+        builder.attributes = Some(attrs);
+
+        let span = tracer.build_with_context(builder, &parent_cx);
+        let cx = parent_cx.with_span(span);
```
paullegranddc, line 68: What about setting span kind to Server

instrumentation/datadog-aws-lambda/src/invocation.rs
```diff
@@ -0,0 +1,265 @@
...
+        cold_start: bool,
+    ) -> Self {
+        let function_name = &lambda_cx.env_config.function_name;
+        let request_id = lambda_cx.request_id.clone();
+
+        tracing::debug!(request_id, "creating invocation root span");
+
+        let parent_cx = Context::current();
+
+        let mut builder = tracer.span_builder(TRACER_NAME);
```
paullegranddc: This should be the operation name I think

instrumentation/datadog-aws-lambda/src/invocation.rs
```diff
@@ -0,0 +1,265 @@
...
+
+        let span = tracer.build_with_context(builder, &parent_cx);
+        let cx = parent_cx.with_span(span);
+
+        Self { cx, request_id }
+    }
+
+    pub(crate) fn set_error(&self, err: &impl std::fmt::Display) {
+        let err_msg = err.to_string();
+        tracing::warn!(request_id = self.request_id, "handler returned error");
```
paullegranddc, line 75: Not sure about logging here 🤔

instrumentation/datadog-aws-lambda/src/lib.rs
```diff
@@ -1,2 +1,223 @@
...
+    }
+
+    fn take_cold_start(&mut self) -> bool {
+        std::mem::replace(&mut self.cold_start, false)
+    }
+}
+
+impl<F, Fut, E, R> Service<LambdaEvent<Box<RawValue>>> for WrappedHandler<F, E, R>
+where
+    F: Fn(LambdaEvent<E>) -> Fut + Send + Sync + 'static,
```
paullegranddc: I would expect the wrapped handler to be a `Service<E>` and not a simple function, if users want to use middlewares after tracing

instrumentation/datadog-aws-lambda/src/lib.rs
```diff
@@ -1,2 +1,223 @@
...
+use serde_json::value::RawValue;
+use std::future::Future;
+use std::marker::PhantomData;
+use std::sync::Arc;
+use std::task::{self, Poll};
+
+type BoxFuture<T> = std::pin::Pin<Box<dyn std::future::Future<Output = T> + Send>>;
+
+#[derive(Default)]
+pub struct Config {
```
paullegranddc: Do we need specific config for this integration or can we just pass a datadog_tracing::Configuration object?

All of the keys here are in the main config.

instrumentation/datadog-aws-lambda/src/lib.rs
```diff
@@ -1,2 +1,223 @@
...
+    /// Service name. Overrides `DD_SERVICE`. Ignored when [`tracing`](Self::tracing) is `Some`.
+    pub service: Option<String>,
+    /// Deployment environment. Overrides `DD_ENV`. Ignored when [`tracing`](Self::tracing) is
+    /// `Some`.
+    pub env: Option<String>,
+    /// Service version. Overrides `DD_VERSION`. Ignored when [`tracing`](Self::tracing) is `Some`.
```
paullegranddc: I think we shouldn't expose the public struct like this. Adding config field here would be a breaking change, so I think we should provide a builder style API

instrumentation/datadog-aws-lambda/src/invocation.rs
Dogbu-cyber: Already set at line 50

instrumentation/datadog-aws-lambda/src/lib.rs
Dogbu-cyber: My thought process was that I wanted to provide a way to set service, env, and version without having to also remember to set client-side stats to false. But I think it's just better to accept a                       datadog_opentelemetry::Config and emphasize that users should set Lambda-specific defaults themselves if they pass in their own.

instrumentation/datadog-aws-lambda/src/lib.rs
Dogbu-cyber: Agreed. I'm still working on how to test this. Soon I will push that commit.

instrumentation/datadog-aws-lambda/src/lib.rs
```diff
@@ -1,2 +1,258 @@
...
+                    let result = invocation.finish(result);
+                    if let Err(err) = provider.force_flush() {
+                        tracing::error!("flush failed: {err}");
+                    }
+                    result
+                });
+            }
+        };
+        let typed_event = LambdaEvent::new(typed_payload, event.context);
+        let fut = inner.call(typed_event);
```
purple4reina: It looks like you're json deserializing the event payload from a string. Is that correct? Why do we need to do this? Shouldn't the event come to us as an object in the first place?

Dogbu-cyber: We could take `LambdaEvent<E>` directly, but that would mean deserialization errors would be handled by the Lambda runtime    before our code runs, meaning an invocation with a payload that couldn't be deserialized into an `E `wouldn't be traced at all.

instrumentation/datadog-aws-lambda/src/lib.rs
```diff
@@ -1,2 +1,258 @@
...
+/// When providing a custom config, set these values yourself to avoid redundant stats
+/// computation or span loss during Lambda freezes.
+///
+/// # Examples
+///
+/// ```ignore
+/// use lambda_runtime::service_fn;
+///
+/// // Zero-config (Lambda defaults applied automatically)
+/// lambda_runtime::run(WrappedHandler::new(service_fn(my_handler), None)).await
```
purple4reina: I'm curious about this and would like to learn more before I form an opinion.

It is my understanding that for a customer handler function `my_handler`, you must wrap it with `service_fn(my_handler)`, then start it with `lambda_runtime::run`.

So, the question is, which wrapper wraps who? Do we wrap the `service_fn` or does the `service_fn` wrap us? I see you've chosen the former. I'm curious to learn more about the benefits/drawbacks of each method and why you think this is the preferred option.

instrumentation/datadog-aws-lambda/src/invocation.rs
```diff
@@ -0,0 +1,259 @@
...
+            KeyValue::new(attr::SPAN_TYPE, "serverless"),
+            KeyValue::new(attr::REQUEST_ID, request_id),
+            KeyValue::new(attr::COLD_START, cold_start),
+            KeyValue::new(attr::FUNCTION_ARN, lambda_cx.invoked_function_arn.clone()),
+            KeyValue::new(attr::FUNCTION_VERSION, lambda_cx.env_config.version.clone()),
+            KeyValue::new(attr::FUNCTION_NAME, function_name.to_lowercase()),
+            KeyValue::new(attr::RESOURCE_NAMES, function_name.clone()),
+            KeyValue::new(attr::DD_ORIGIN, "lambda"),
+        ];
+        builder.attributes = Some(attrs);
```
purple4reina, line 65: Note to self, attrs on aws.lambda spans in other languages:
+ [extension](https://github.com/DataDog/datadog-lambda-extension/blob/4134b07e2ac03db548590184077e2203fc49833c/bottlecap/src/lifecycle/invocation/processor.rs#L174-L179)
+ [python](https://github.com/DataDog/datadog-lambda-python/blob/3e094174172e641e0ddc7581af9c5924334915e5/datadog_lambda/tracing.py#L1458-L1471)
+ [node](https://github.com/DataDog/datadog-lambda-js/blob/c9b398b8415fd33f857a215bc75af991a22e67c4/src/index.ts#L466-L476)

purple4reina: Looks like one thing we add in other languages is the version of the instrumentation library. Any chance we can add the version of this rust crate here as a tag as well?

instrumentation/datadog-aws-lambda/src/lib.rs
purple4reina: Okay, that makes sense. What I wonder though is, how often is there going to be an error deserializing event payloads? Probably rarely if never. Were you ever able to produce an error during deserialization? If so, how were you able to? I suspect it's not possible.

instrumentation/datadog-aws-lambda/src/lib.rs
Dogbu-cyber: `WrappedHandler` is a `Service` that wraps an inner `Service`. which means we can place tower middleware between tracing and the handler:

```rust
let inner = ServiceBuilder::new()
    .layer(auth_middleware)
    .service(service_fn(my_handler));

lambda_runtime::run(WrappedHandler::new(inner, None)).await
```


If we were to instead let service_fn wrap us, my understanding is the api would look something like this

```rust
let traced = datadog_lambda::wrap(my_handler);
lambda_runtime::run(service_fn(traced)).await
```

`wrap` returns a new async function that starts the span, calls the original handler, ends
the span, and flushes. The API is simpler, but because the result is a function, not a `Service`, you can't insert tower middleware between tracing and the handler:

```rust
// wrap(my_handler) is a function, not a Service. So this is not possible
let traced = datadog_lambda::wrap(my_handler);
let inner = ServiceBuilder::new()
    .layer(auth_middleware)
    .service(???); // traced is a fn, not a Service
```

This is the sole reason I went with the code as it's currently written.

purple4reina: what's tower middleware and why is it important that we support it?

purple4reina: ooooh, I see.

instrumentation/datadog-aws-lambda/src/lib.rs
Dogbu-cyber: It could happen if someone invoked a Lambda directly with a custom payload that doesn't match the handler's expected type. 

**A much bigger reason I forgot to mention though is looking ahead at #190 .**

Let's say we accept a `LambdaEvent<E>`. 

The lambda runtime will deserialize raw bytes into that E, but now we need to perform trace extraction.

The problem is the inferrer has no clue what E is (since it's designed to in the future to be generic across runtimes), it only knows how to interpret a json payload and match a trigger from that point, which is why the options are to either accept a `Value,` or a `RawValue`.

purple4reina: Okay, that's helpful. And the crate that you and Jordan were working on for trace extraction. It accepts an event. What is the type of that event? Does it take a string?

Dogbu-cyber: These are the 3 methods exposed by the trace-inferrer crate for trace extraction:

`pub fn infer_span_from_bytes(&self, payload: &[u8]) `
`pub fn infer_span_from_value(&self, payload: &Value)`
`pub fn infer_span(&self, payload: &str)`

Also to clarify, I didn't work on this crate.

instrumentation/datadog-aws-lambda/src/lib.rs
paullegranddc: You can just ask them to pass a `datadog_opentelemetry::ConfigBuilder` and then override the specific config fields

Dogbu-cyber:
`datadog-opentelemetry` is pulled in with `features = ["test-utils"]` as a production dependency because `set_trace_writer_synchronous_write` is gated behind `#[cfg(feature = "test-utils")]` upstream. Once that setter is ungated,  the `test-utils` feature from `[dependencies]` should be dropped to avoid inflating the binary.

instrumentation/datadog-aws-lambda/src/lib.rs
Dogbu-cyber: Done

purple4reina (approved):

instrumentation/datadog-aws-lambda/src/lib.rs
```diff
@@ -1,2 +1,262 @@
...
+        std::mem::replace(&mut self.cold_start, false)
+    }
+}
+
+impl<S, E, R> Service<LambdaEvent<Box<RawValue>>> for WrappedHandler<S, E, R>
+where
+    S: Service<LambdaEvent<E>, Response = R, Error = lambda_runtime::Error>
+        + Clone
+        + Send
+        + 'static,
```
paullegranddc: I think the bound on the error should be something like 
`S::Error: Into<Diagnostic> + Debug, instead of lambda_runtime::Error, to match the signature of [lambda_runtime::run](https://docs.rs/lambda_runtime/latest/lambda_runtime/fn.run.html)

instrumentation/datadog-aws-lambda/src/lib.rs
```diff
@@ -1,2 +1,262 @@
...
+            }
+        };
+        let typed_event = LambdaEvent::new(typed_payload, event.context);
+        let fut = inner.call(typed_event);
+        Box::pin(async move {
+            let result = fut.with_context(invocation.handler_context()).await;
+            let result = invocation.finish(result);
+            if let Err(err) = provider.force_flush() {
+                tracing::error!("flush failed: {err}");
+            }
```
paullegranddc: this should not be necessary if the tracer is in synchronous mode

instrumentation/datadog-aws-lambda/src/lib.rs
```diff
@@ -1,2 +1,262 @@
...
+                    let result = invocation.finish(result);
+                    if let Err(err) = provider.force_flush() {
+                        tracing::error!("flush failed: {err}");
+                    }
+                    result
+                });
+            }
+        };
+        let typed_event = LambdaEvent::new(typed_payload, event.context);
+        let fut = inner.call(typed_event);
```
paullegranddc: `inner.call` should be wrapper in the `invocation.handler_context()`
```suggestion
        let fut = {
                let _ctx = invocation.handler_context().attach();
                inner.call(typed_event)
          };
```

paullegranddc (approved):

instrumentation/datadog-aws-lambda/src/attribute_keys.rs
```diff
@@ -0,0 +1,18 @@
+// Copyright 2025-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+pub(crate) const OPERATION_NAME: &str = "operation.name";
```
purple4reina, line 4: TODO: We need to double check this. When this is `operation_name` then the invocation doesn't show up on the serverless page.

instrumentation/datadog-aws-lambda/src/lib.rs
ygree: Address this in 70f982c0d40e85056aa090b9a3478c6ba728470a

instrumentation/datadog-aws-lambda/src/invocation.rs
ygree: Fixed. BTW, why do we need both the span name and the operation attribute?

instrumentation/datadog-aws-lambda/src/lib.rs
ygree: I have changed it so that it uses `datadog_opentelemetry::ConfigBuilder` directly. Therefore, there is no need for a separate Config structure for now. See 2dbf5520aaeb15deb0dae35e7dff2314915d2f0f

instrumentation/datadog-aws-lambda/src/lib.rs
ygree: I have changed it so that it uses datadog_opentelemetry::ConfigBuilder directly. Therefore, there is no need for a separate Config structure for now. See https://github.com/DataDog/dd-trace-rs/commit/19feab9d3695d1f2f40dc84a56068b7c42811860

instrumentation/datadog-aws-lambda/src/lib.rs
ygree: Fixed in b9e8297bdb568d83410b570388285134a5a6fab5

instrumentation/datadog-aws-lambda/src/lib.rs
ygree: Fixed in aa5fc55afc24f7edb9dcad71e3929c49f68747d8

instrumentation/datadog-aws-lambda/src/lib.rs
ygree: Thanks! Fixed in 817e9eaf75a7bf48d713a6cc9ef544df5e9fee69. to ensure that the sync part is within the same invocation span and capture the same active context for the future execution.

ygree:
> `datadog-opentelemetry` is pulled in with `features = ["test-utils"]` as a production dependency because `set_trace_writer_synchronous_write` is gated behind `#[cfg(feature = "test-utils")]` upstream. Once that setter is ungated, the `test-utils` feature from `[dependencies]` should be dropped to avoid inflating the binary.

Added a TODO into `Cargo.toml` and cleaned up flush calls in aa5fc55afc24f7edb9dcad71e3929c49f68747d8
Ref #221

instrumentation/datadog-aws-lambda/Cargo.toml
```diff
@@ -13,16 +13,17 @@ authors.workspace = true
 publish.workspace = true
 
 [dependencies]
-base64 = "0.22"
-datadog-opentelemetry = { version = "0.3", path = "../../datadog-opentelemetry" }
+# TODO: Drop `test-utils` from this production dependency once
+# `ConfigBuilder::set_trace_writer_synchronous_write` is ungated upstream.
+datadog-opentelemetry = { version = "0.3", path = "../../datadog-opentelemetry", features = ["test-utils"] }
```
purple4reina: Will we need to remove the `path` before releasing this?

purple4reina (approved):

instrumentation/datadog-aws-lambda/Cargo.toml
ygree: There is no need to remove the `path`. Cargo uses the local path for development and the version when this crate is published. The `version` will also remain at `0.3`, meaning it's the latest published version in the `[0.3.0, 0.4.0)` interval.

ygree:
> `datadog-opentelemetry` is pulled in with `features = ["test-utils"]` as a production dependency because `set_trace_writer_synchronous_write` is gated behind `#[cfg(feature = "test-utils")]` upstream. Once that setter is ungated, the `test-utils` feature from `[dependencies]` should be dropped to avoid inflating the binary.

This was addressed once #221 was merged into the main branch.

PR #214, opened 2026-04-27
fix(rc): cached files hash

iunanua:
# What does this PR do?

- Use the decoded file bytes instead of the base64 bytes to calculate SHA256 hash.
- Remove an allocation and use directly the decoded bytes in `process_config` method

# Motivation

By creating the hash from the base64 bytes, the agent is unable to confirm that a config has been applied and keeps sending it constantly.


bantonsson (approved):
Nice catch

PR #216, opened 2026-04-30
feat(telemetry): propagate session IDs and process lineage to libdatadog

mabdinur:
## Background

An app often runs as a tree of processes — workers, subcommands, a Lambda re-exec'ing. Each process initializes its own tracer with a fresh `runtime_id`, so without extra signal the backend sees them as unrelated. This PR tags every telemetry event with `session_id`, `parent_session_id`, and `root_session_id` so the backend can stitch them into one logical instrumentation session.

No new public API surface — all wiring is `pub(crate)`.

## Wire protocol

Two env vars carry lineage between processes:

- `_DD_PARENT_RS_SESSION_ID` — immediate parent's `runtime_id`
- `_DD_ROOT_RS_SESSION_ID` — `runtime_id` of the root of the spawn chain

At library load (see **Why `ctor` and `libc`?** below), a one-shot hook:

1. **Captures inbound** — reads the env vars set by the spawning process and builds `TelemetryInstrumentationSessions { session_id = runtime_id, root, parent }`. Values matching `runtime_id` are dropped.
2. **Installs outbound** — writes `_DD_PARENT_RS_SESSION_ID = runtime_id`, and `_DD_ROOT_RS_SESSION_ID = runtime_id` if it was unset. Subsequent `Command::spawn` inherits these automatically.

Capture runs before install so we don't reflect our own id back as our parent. The hook also calls `Config::process_runtime_id()` so the process `runtime_id` is fixed at the same early point.

Later, `sessions_from_runtime_id` returns the cached value (or falls back to a fresh env read) and passes it to libdatadog from two init paths:

- **`core/telemetry.rs`** — sets `session_id` / `root_session_id` / `parent_session_id` on the libdd-telemetry worker config
- **`span_exporter.rs`** — calls `set_telemetry_instrumentation_sessions` on the trace exporter builder (alongside existing `TelemetryConfig.runtime_id`)

## Why `ctor` and `libc`?

This PR adds `ctor = "0.2"` and moves `libc = "0.2"` from a Linux-only target dependency to a general dependency in `datadog-opentelemetry/Cargo.toml`.

**`ctor`** registers `install_lineage()` as a dynamic-library constructor (`.init_array` on Linux/macOS) via `#[ctor::ctor]`. That runs **between `exec` and `main`**, before user threads exist — the window where `libc::setenv` is POSIX-safe.

We need that early hook because tracer init (`DatadogTracingBuilder::init`) happens too late on its own:

- **Inbound** lineage should be captured as early as possible.
- **Outbound** vars must be set before the app spawns subprocesses, including spawns that happen before or outside explicit tracer init.

Without `ctor`, lineage would only be read when something first called `sessions_from_runtime_id`; outbound env would not be installed for early `Command::spawn` calls.

**`libc`** is used directly in the constructor for `getenv` / `setenv` on Unix and `putenv_s` on Windows, matching the module's threading safety contract.

**Fallback:** on targets without `.init_array` support, `sessions_from_runtime_id` still works via `std::env::var` — lineage degrades gracefully (no early outbound install, no panic).

**Lockfile impact vs `main`:** only `ctor` is added as a new resolved package in `Cargo.lock` and `instrumentation/Cargo.lock` (`+11` lines each). `libc` was already present transitively on `main`.

## Env scope

`setenv` writes to **this process's** env block — there is no host-wide registry. Effects:

- **Inherited** by children spawned after the ctor hook runs (env is snapshotted at spawn).
- **Not visible** to sibling processes, already-running processes, other shells/containers/users, or the parent shell after we exit.
- **Shared with threads** in our own process; the ctor hook runs before any user thread is spawned.
- **Broken by `Command::env_clear` / `env_remove`.** If any process in the chain wipes its env before spawning (common in sandboxes, hermetic builders, container runtimes), the child starts a new lineage tree. Graceful degradation: no panic, just a missed correlation.

## Caveats

**Threading.** `setenv` is unsound on POSIX when raced with `getenv` from another thread. The `#[ctor::ctor]` hook runs before user threads exist, which is why env mutation is done there rather than during tracer init.

**Fork.** Lineage is correct across `fork+exec` (the default for `Command::spawn`) because `exec` resets memory and the constructor re-runs. Bare `fork()` (Unix daemons, `nix::unistd::fork`) is **not** handled: the child inherits both the parent's `Config::process_runtime_id()` `OnceLock` and our `CAPTURED` `OnceLock`, so it reports the parent's session. A fix requires resetting both from a `pthread_atfork` child handler — out of scope here; noted in a module comment.

## Implementation

- `core/telemetry_session.rs` (new, `pub(crate)`): `#[ctor::ctor]` capture-and-install; pure `sessions_from_env` unit-tested with an injected env closure.
- `core/configuration/configuration.rs`: `process_runtime_id()` made `pub(crate)` so the ctor hook can share the same runtime id as `Config`.
- `core/telemetry.rs` and `span_exporter.rs`: consume `sessions_from_runtime_id` at init (see Wire protocol).

## Test plan

- [x] `telemetry_session::` unit tests (3)
- [x] `cargo fmt --check`, `cargo hack -p datadog-opentelemetry --each-feature clippy`
- [x] `cargo check --workspace --locked` (Rust 1.87.0)
- [x] `instrumentation`: `cargo test --workspace --locked` (stable, AWS integration CI)


mabdinur:
Blocked by the release: https://github.com/DataDog/libdatadog/pull/1822

mabdinur:
## Next steps before merging

This PR currently carries `[patch.crates-io]` git overrides for six libdatadog crates because the APIs it depends on (`TelemetryInstrumentationSessions`, `NativeCapabilities`, `session_id` fields on telemetry `Config`) were added in [DataDog/libdatadog#1822](https://github.com/DataDog/libdatadog/pull/1822) without incrementing the crate version numbers, so they cannot be consumed directly from crates.io.

**The patch overrides have now been dropped in the latest commit**, pointing instead at the versions that will exist once [DataDog/libdatadog#1989](https://github.com/DataDog/libdatadog/pull/1989) merges and the crates are published:

| Crate | Before | After |
|---|---|---|
| `libdd-data-pipeline` | `3.0.1` (git patch) | `3.1.0` |
| `libdd-telemetry` | `4.0.0` (git patch) | `5.0.0` |
| `libdd-capabilities-impl` | `1.0.0` (git patch) | `1.1.0` |

**This PR will not build until the following sequence completes:**

1. ✅ **[DataDog/libdatadog#1989](https://github.com/DataDog/libdatadog/pull/1989)** — bumps `libdd-capabilities-impl` to `1.1.0` and `libdd-data-pipeline` to `3.1.0` (already open)
2. ⏳ **libdatadog publishes** `libdd-capabilities-impl 1.1.0`, `libdd-data-pipeline 3.1.0`, and `libdd-telemetry 5.0.0` to crates.io
3. ⏳ **Regenerate lock files** (`Cargo.lock` and `instrumentation/Cargo.lock`) against the new published versions
4. ⏳ **Merge this PR** — no system-tests changes required once the crates are published

Cargo.toml
```diff
@@ -23,11 +23,15 @@ authors = ["Datadog Inc. <info@datadoghq.com>"]
 
 [workspace.dependencies]
 # Libdatadog dependencies - change to a stable version once we release
-libdd-data-pipeline = { version = "3.0.1", default-features = false }
-libdd-trace-utils = { version = "3.0.0", default-features = false }
-libdd-telemetry = { version = "4.0.0", default-features = false }
-libdd-common = { version = "3.0.1", default-features = false }
-libdd-tinybytes = { version = "1.1.0", default-features = false }
+libdd-data-pipeline = { version = "4.0.0", default-features = false, features = [
+    "https",
```
iunanua: do you need `https` or it has slipped in?
if you need it might be best to add a new `https` feature that is off by default.

datadog-opentelemetry/src/exporter/mod.rs
```diff
@@ -607,10 +610,16 @@ impl<T: Send + 'static> TraceExporterWorker<T> {
...
+            // Block on the async `wait_agent_info_ready` from this plain
+            // `thread::spawn`; needed for deterministic sampling in tests.
+            tokio::runtime::Builder::new_current_thread()
+                .enable_all()
+                .build()
+                .expect("tokio runtime for wait_agent_info_ready")
+                .block_on(
+                    self.trace_exporter
+                        .wait_agent_info_ready(Duration::from_secs(5)),
+                )
```
paullegranddc: This conflicts with https://github.com/DataDog/dd-trace-rs/pull/226, you should wait for this to be merged before rebasing this PR

datadog-opentelemetry/src/core/telemetry_session.rs
```diff
@@ -0,0 +1,111 @@
...
+// Callers must invoke `DatadogTracingBuilder::init` early in `main` before
+// spawning threads; `OnceLock` makes the install one-shot.
+#[allow(clippy::disallowed_methods)]
+fn install_outbound_env(runtime_id: &str) {
+    // Preserve root from an upstream parent; otherwise we become the root.
+    if std::env::var_os(ENV_ROOT_RS_SESSION_ID).is_none() {
+        std::env::set_var(ENV_ROOT_RS_SESSION_ID, runtime_id);
+    }
+    std::env::set_var(ENV_PARENT_RS_SESSION_ID, runtime_id);
+}
```
paullegranddc: This is still not great as there is no guarantee that this is called early enough.
I would prefer if you used the [ctor crate](https://docs.rs/ctor/latest/ctor/) to verify that this is called whenever the process starts.
The ctor function should also call libc::setenv and getenv instead of the rust stdlib https://docs.rs/libc/latest/libc/fn.setenv.html

datadog-opentelemetry/src/core/telemetry_session.rs
```diff
@@ -0,0 +1,111 @@
+// Copyright 2025-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+//! Telemetry session ids. First call captures inbound lineage from
+//! `_DD_ROOT_RS_SESSION_ID` / `_DD_PARENT_RS_SESSION_ID` and installs the
+//! outbound env so subprocesses inherit it via `Command::spawn`.
+//!
+//! Best-effort: not refreshed on bare `fork()`. The child shares the parent's
```
paullegranddc: We could do something on fork using the `pthread_atfork` libc call, but rust apps in general don't continue after forking because it's quite dangerous

datadog-opentelemetry/src/exporter/mod.rs
mabdinur: Sounds good, will wait. Once SharedRuntime is in I can drop the block_on block and use it directly, which should zero out the exporter diff. Do you have a rough ETA on #226?

datadog-opentelemetry/src/core/telemetry_session.rs
mabdinur: Done in 5a246a30

datadog-opentelemetry/src/core/telemetry_session.rs
mabdinur: I left it as a doc note for now. Is this something we should handle in this PR? I don't think this concern is addressed anywhere? I'd prefer to defer it

Cargo.toml
mabdinur: The feature was added as part of the libdatadog 3.x to 4.x bump. 

The feature gates TLS support and without it DD_TRACE_AGENT_URL=https://... configs would silently lose HTTPS. Since that's a valid config users can set today, leaving it on by default seems like the safe default.

iunanua: but it has the drawback that it requires a lot of dependencies :/

wdyt @paullegranddc @bantonsson?

mabdinur: I reverted this change. This PR bumped `libdd-data-pipeline` without adding the https feature: https://github.com/DataDog/dd-trace-rs/pull/236/changes#diff-47e885b39da0467fd3ae1ceeddf92f0cdfb09610fb1744d69e76648055270e24.

datadog-opentelemetry/src/exporter/mod.rs
iunanua: `libdd-data-pipeline` was bumped [here](https://github.com/DataDog/dd-trace-rs/pull/236/changes)
I guess we could merge this PR now 
@paullegranddc @Aaalibaba42 wdyt?

khanayan123 (approved):

iunanua (approved):

PR #217, opened 2026-04-30
feat(baggage): fix DatadogPropagator::fields

iunanua:
# What does this PR do?

- change `DatadogCompositePropagator` to obtain keys from injectors according to [Otel spec](https://opentelemetry.io/docs/specs/otel/context/api-propagators/#fields-1).
- remove the not needed `DatadogPropagator::fields` property

### DIFF
--- a/datadog-opentelemetry/src/text_map_propagator.rs
+++ b/datadog-opentelemetry/src/text_map_propagator.rs
@@ -69,7 +69,6 @@ pub struct DatadogPropagator {
     baggage_propagator: BaggagePropagator,
     baggage_extract: bool,
     baggage_inject: bool,
-    fields: Vec<String>,
 }
 
 impl DatadogPropagator {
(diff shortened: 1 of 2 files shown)


paullegranddc (commented):
Are you sure this is what's described in the spec?
Because from what I see, thje `Keys` section is under the `TextMap Extract` title.

The doc specific to B3 headers mentions 
```Fields
Fields MUST return the header names that correspond to the configured format, i.e., the headers used for the inject operation
```
 but it seems very specific to B3 🤔

datadog-opentelemetry/src/text_map_propagator.rs
```diff
@@ -238,7 +224,7 @@ impl TextMapPropagator for DatadogPropagator {
 
     fn fields(&self) -> opentelemetry::propagation::text_map_propagator::FieldIter<'_> {
         let fields: &[String] = if self.cfg.enabled() {
-            &self.fields
+            self.inner.keys()
```
bantonsson: This is soo confusing... is it `keys` or `fields`? They are not always the same.

iunanua:
> Are you sure this is what's described in the spec? Because from what I see, thje `Keys` section is under the `TextMap Extract` title.
> 
> The doc specific to B3 headers mentions
> 
> ```
> Fields MUST return the header names that correspond to the configured format, i.e., the headers used for the inject operation
> ```
> 
> but it seems very specific to B3 🤔

There is some confusion about this:
Otel `TextMapPropagator` trait defines the `fields()` method. It returns an iter of fields used by [`TextMapPropagator`] 
and [spec says](https://opentelemetry.io/docs/specs/otel/context/api-propagators/#fields-1)
> Fields MUST return the header names that correspond to the configured format, i.e., the headers used for the inject operation.

But our `DatadogCompositePropagator` has the `keys()` method. It returned the header keys of the configured extractors which were ultimately used as values for the otel fields...

datadog-opentelemetry/src/text_map_propagator.rs
iunanua: Indeed. 
I'm going to rename the `keys()` as `fields()`

bantonsson:
@paullegranddc With regards to what fields should return, the relevant part of the generic [spec is here](https://opentelemetry.io/docs/specs/otel/context/api-propagators/#fields) and it says:
> The predefined propagation fields. If your carrier is reused, you should delete the fields here before calling [Inject](https://opentelemetry.io/docs/specs/otel/context/api-propagators/#inject).

bantonsson (approved):

PR #218, opened 2026-05-04
feat(propagation): limit the number of keys parsed from tracestate

paullegranddc:
# Motivation

We want to limit the amount of keys we allow when parsing a tracestate

# Changes

* Move TraceState, TraceParent and InjectTraceState to the tracecontext module as this is where they belong
* Take only the first 32 entries of tracestate when parsing it.


datadog-opentelemetry/src/propagation/tracecontext.rs
```diff
@@ -252,7 +447,7 @@ fn inject_tracestate(context: &InjectSpanContext, carrier: &mut dyn Injector) {
 
     // Add additional tracestate values if present
     if let Some(ts) = &context.tracestate {
-        for part in ts.additional_values().take(31) {
+        for part in ts.additional_values().take(TRACESTATE_MAX_MEMBERS - 1) {
```
ekump, line 450: should we add a regression test where we add more than `TRACESTATE_MAX_MEMBERS`?

ekump (approved):
a unit test would be nice for 32+ members, but LGTM

PR #223, opened 2026-05-08
feat(baggage): add support for baggage span tags

rachelyangdog:
# What does this PR do?

Implements the `DD_TRACE_BAGGAGE_TAG_KEYS` feature, which automatically promotes W3C Baggage entries to span tags at extraction time.

When a span is created with a remote parent, baggage entries matching the configured filter are attached as span attributes prefixed with `baggage.`: 
> baggage: user.id=alice,session.id=xyz123                                                                                                                                                                           
>   → span tag baggage.user.id = "alice"                                                                                                                                                                               
>   → span tag baggage.session.id = "xyz123"                                                                                                                               

Tags are only applied at extraction time, not to locally-created root spans.

Baggage Tagging Behavior
- `""` (empty): Disabled — no baggage keys become span tags
- `"*"`: All baggage keys are tagged
- `"user.id,session.id"`: Only the listed keys are tagged
- (unset): Default is user.id, session.id, account.id

**W3C spec compliance fix**              
                                                                                                                                                                             Per the W3C Baggage spec, a malformed member (empty key, empty value, or missing =) must cause all subsequent members to be dropped — not just the malformed one. This PR fixes the extraction to use stop-at-first-error semantics instead of skip-and-continue.
- Before: user.id=,session.id=abc → {session.id: "abc"} (partial extraction)                                                                                                                                         
- After: user.id=,session.id=abc → {} (entire header dropped after malformed entry)

**Configuration**                                                                                                                     
- Env var: `DD_TRACE_BAGGAGE_TAG_KEYS`                      
- Default: "user.id,session.id,account.id"                                                                                                                       

# Motivation

# Additional Notes


datadog-opentelemetry/src/propagation/baggage.rs
```diff
@@ -52,6 +52,13 @@ fn parse_baggage_member(baggage_member: &str) -> Option<KeyValueMetadata> {
         return None;
     };
 
+    let name = name.trim();
+    let value = value.trim();
+    if name.is_empty() {
```
iunanua, line 57: According to the PR description

> Per the W3C Baggage spec, a malformed member (empty key, empty value, or missing =) must cause all subsequent members to be dropped

is an empty value considered malformed?

datadog-opentelemetry/src/core/configuration/configuration.rs
```diff
@@ -807,6 +807,48 @@ impl Display for ServiceName {
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let trimmed = s.trim();
+        if trimmed.is_empty() {
+            Ok(BaggageTagKeyFilter::Disabled)
+        } else if trimmed == "*" {
+            Ok(BaggageTagKeyFilter::All)
+        } else {
+            Ok(BaggageTagKeyFilter::Keys(
+                trimmed
```
iunanua: should it return  `BaggageTagKeyFilter::Disabled` if  `trimmed.split(',').map(|k| k.trim().to_string()).filter(|k| !k.is_empty()).collect()` returns an empty list?

datadog-opentelemetry/src/propagation/baggage.rs
```diff
@@ -144,16 +153,6 @@ mod tests {
                 .into_iter()
                 .collect(),
             ),
-            // "valid header with no value"
```
iunanua, line 147: This test was deleted but it seems that the new code produces the same result (since an empty value is still valid)

datadog-opentelemetry/src/propagation/baggage.rs
iunanua: aah, I just saw you latest commit. So I guess the PR description is wrong and empty values are valid

rachelyangdog: Yes, the PR description is outdated. The W3C spec defines value = baggage-octet (zero or more octets), so empty values are valid. Only an empty key is considered malformed since key = 1tchar requires at least one character. I've updated the code and the test accordingly.

datadog-opentelemetry/src/core/configuration/configuration.rs
rachelyangdog: Yes, good catch. Fixed with a test in commit [0a0c4fe](https://github.com/DataDog/dd-trace-rs/pull/223/commits/0a0c4fe77c24f9ee6c77f645414efbed75e50104)!

datadog-opentelemetry/src/propagation/baggage.rs
rachelyangdog: Restored!

iunanua:
A ST is failing because it expects all items get dropped if a value is missing :upside_down_face:

iunanua (approved):

PR #225, opened 2026-05-12
feat(client-stats): expose css obfuscation param

Eldolfin:
# What does this PR do?

Expose the experimental client side stats obfuscation for internal testing/benchmarking.

# Motivation

# Additional Notes

TODO:
- [x] put back actual crate versions when libdatadog gets bumped

APMSP-2969


Cargo.lock
```diff
@@ -1312,6 +1312,7 @@ dependencies = [
  "libdd-shared-runtime",
  "libdd-telemetry",
  "libdd-tinybytes",
+ "libdd-trace-obfuscation",
```
iunanua, line 1285: As `stats-obfuscation` is being added to the `libdd-data-pipeline` features, `libdd-trace-obfuscation` is always pulled.
Should we add an `experimental-stats-obfuscation` (or equivalent) feature so production users don't pay for it? Or is the dep small that it's not worth the trouble?

Eldolfin: I initially wanted to exclude `libdd-trace-obfuscation` too, but a CSS fix (https://github.com/DataDog/libdatadog/pull/1944) made it a required dependency of `libdd-trace-stats` (via `libdd-data-pipeline`), so it's included in the final build anyway.

iunanua: there's no way out

iunanua (approved):

PR #227, opened 2026-05-15
fix(rc): accept RC list-shape tags and honor tracing_sampling_rate

bm1549:
# What does this PR do?

Fixes adaptive-sampling Remote Config in `datadog-opentelemetry`, rebuilt on top of PR #154. Also adds `DD_TRACE_SAMPLE_RATE` support.

Before this PR, four things were broken:

1. **`tracing_sampling_rate` from RC was ignored.** The handler only acted on `tracing_sampling_rules`; a rate-only payload installed nothing.
2. **RC list-shape `tags` were rejected.** RC encodes `tags` as `[{"key": "env", "value_glob": "prod"}]`, but `libdd_sampling::SamplingRuleConfig::tags` only accepted the map shape, so the parse errored and the whole update was dropped.
3. **Env `DD_TRACE_SAMPLING_RULES` were wiped on every RC update.** `update_sampling_rules_from_remote` does a full override, so any RC change replaced env rules, even when RC only sent a global rate.
4. **`DD_TRACE_SAMPLE_RATE` had no effect.** No binding existed.

# Motivation

End-to-end adaptive sampling didn't work.

# What changed

**Composition.** `ApmTracingHandler::process_config` now follows the multi-source precedence model:

| env rules | env `DD_TRACE_SAMPLE_RATE` | RC `tracing_sampling_rules` | RC `tracing_sampling_rate` | Effective rule chain |
|---|---|---|---|---|
| present | any | absent / null | absent / null | `env_rules` |
| present | unset | absent / null | present | `env_rules + catch_all(rc_rate)` |
| present | set | absent / null | absent / null | `env_rules + catch_all(env_rate)` |
| present | set | absent / null | present | `env_rules + catch_all(rc_rate)` |
| any | any | non-empty array | absent / null | `rc_rules + catch_all(env_rate)` if env_rate set |
| any | any | non-empty array | present | `rc_rules + catch_all(rc_rate)` |

The synthetic catch-all uses libdatadog's default provenance, mapping to DM `-3` (LOCAL_USER). See the "legacy behavior" comment in [test_trace_sampling_rules_override_rate](https://github.com/DataDog/system-tests/blob/main/tests/parametric/test_dynamic_configuration.py#L872).

**`DD_TRACE_SAMPLE_RATE`.** New `Config::trace_sample_rate(): Option<f64>`. When set, the sampler installs an implicit catch-all so `DD_TRACE_RATE_LIMIT` applies. Unset means no catch-all (libdatadog's no-rule path samples at 100%).

**Tag normalization.** RC encodes `tags` as the list shape `[{"key", "value_glob"}]`. This is parsed natively by `libdd-sampling` ≥ 2.1.0 (DataDog/libdatadog#2033), so this PR bumps `libdd-sampling` 1.0.0 → 2.1.0 (pulling `libdd-common` → 4.2.0) and no in-tracer normalization is needed. An earlier revision carried a `normalize_rc_tags` shim for this; it has been removed now that the upstream release is available. Regression coverage: `test_handler_rc_rules_with_list_tags_applied` (list-shape tags apply, tags preserved as a map) and `test_handler_malformed_tags_rejects_update` (malformed list entries still rejected wholesale, not silently broadened).

**Fail-closed behavior.** When libdatadog rejects an update (malformed tags, out-of-range rate), `process_config` returns `Err` so the RC dispatcher reports `apply_state=3` and the prior policy survives. Out-of-range RC `tracing_sampling_rate` (outside `[0.0, 1.0]`) and non-numeric values are rejected up-front.

**Env/code rate validation.** `DD_TRACE_SAMPLE_RATE` (and the programmatic `set_trace_sample_rate`) get the same range check: only finite values in `[0.0, 1.0]` are honored; out-of-range values are logged and treated as unset rather than installed as a catch-all rule that libdatadog would clamp (negative ⇒ drop all, > 1.0 ⇒ keep all).

**Target check.** A config's `service_target` is honored before applying: a payload whose specific (non-`*`) `service`/`env` doesn't match this tracer — primary service or an advertised extra service, compared case-insensitively — is ignored, so a mistargeted RC delivery can never change this service's sampling. Mirrors dd-trace-py/go.

# Additional Notes

- Four parametric tests unblocked by this PR. Companion PR DataDog/system-tests#7007.
- Coordinated with @iunanua's PR #222 (libdatadog RC client wiring). #227 lands first; #222 rebases on top.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ libdd-data-pipeline = { version = "5.0.0", features = [
 libdd-trace-utils = { version = "5.0.0", default-features = false }
 libdd-capabilities-impl = { version = "2.0.0", default-features = false }
 libdd-telemetry = { version = "5.0.0", default-features = false }
-libdd-common = { version = "4.1.0", default-features = false }
+libdd-common = { version = "4.2.0", default-features = false }
 libdd-tinybytes = { version = "1.1.1", default-features = false }
 libdd-library-config = { version = "2.0.0", default-features = false }
 libdd-sampling = { version = "2.1.0", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ libdd-telemetry = { version = "5.0.0", default-features = false }
 libdd-common = { version = "4.1.0", default-features = false }
 libdd-tinybytes = { version = "1.1.1", default-features = false }
 libdd-library-config = { version = "2.0.0", default-features = false }
-libdd-sampling = { version = "1.0.0", default-features = false }
+libdd-sampling = { version = "2.1.0", default-features = false }
 opentelemetry_sdk = { version = "0.31.0", features = [
     "trace",
     "metrics",
--- a/datadog-opentelemetry/src/core/configuration/configuration.rs
+++ b/datadog-opentelemetry/src/core/configuration/configuration.rs
@@ -1712,11 +1712,13 @@ impl Config {
     /// active Remote Config override: locally-configured rules followed by an
     /// implicit catch-all that applies `DD_TRACE_SAMPLE_RATE`.
     ///
-    /// The catch-all is appended only when the env rate is finite and not the
-    /// default of 1.0 — a 1.0 catch-all would match every unmatched span at
-    /// 100% sampling, which is identical to having no catch-all (libdatadog's
-    /// fallback path samples at 1.0 when no rule matches), so we omit it to
-    /// avoid noisy `_dd.p.dm` tags on the wire.
+    /// The catch-all is appended only when `DD_TRACE_SAMPLE_RATE` is explicitly
+    /// set. Its default is unset (`None`), in which case nothing is appended and
+    /// libdatadog's no-rule fallback samples unmatched spans at 100%. An explicit
+    /// value — including `1.0` — does install the catch-all, so `DD_TRACE_RATE_LIMIT`
+    /// applies to otherwise-unmatched spans. The rate is already validated to a
+    /// finite value in `[0.0, 1.0]` at ingestion; the `is_finite` guard below is
+    /// belt-and-suspenders.
     pub(crate) fn effective_initial_rules(&self) -> Vec<libdd_sampling::SamplingRuleConfig> {
         let mut rules: Vec<libdd_sampling::SamplingRuleConfig> = self
             .local_trace_sampling_rules()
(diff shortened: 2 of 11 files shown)


iunanua (approved):
LGTM

I guess we can merge it before #222 because it will take a little while before we fully resolve it

BTW I have tested it with a real agent but I'm not able to see RC payloads when I enable adaptive sampling in the backend 🤷

datadog-opentelemetry/src/propagation/baggage.rs
```diff
@@ -228,7 +228,7 @@ mod tests {
             assert_eq!(baggage.len(), expected_keys.len(), "header: {header:?}");
             for key in expected_keys {
                 assert!(
-                    baggage.get(&Key::new(key)).is_some(),
+                    baggage.get(Key::new(key)).is_some(),
```
bm1549, line 231: This is an unrelated clippy fix on Rust 1.95.0 for `clippy::needless_borrows_for_generic_args`

datadog-opentelemetry/src/core/configuration/configuration.rs
```diff
@@ -1639,17 +1708,45 @@ impl Config {
...
-
-        // Fallback rules are locally defined, so "local" provenance is correct
-        let internal: Vec<libdd_sampling::SamplingRuleConfig> = self
-            .trace_sampling_rules()
+    /// Composes the rules that the sampler should see in the absence of any
+    /// active Remote Config override: locally-configured rules followed by an
+    /// implicit catch-all that applies `DD_TRACE_SAMPLE_RATE`.
+    ///
+    /// The catch-all is appended only when the env rate is finite and not the
+    /// default of 1.0 — a 1.0 catch-all would match every unmatched span at
```
iunanua: Is _'and not the default of 1.0'_ comment still valid?
Now the default value for `trace_sample_rate` is None, no?

Cargo.toml
```diff
@@ -32,7 +32,7 @@ libdd-telemetry = { version = "5.0.0", default-features = false }
 libdd-common = { version = "4.1.0", default-features = false }
```
iunanua: `libdd-sampling@2.1.0` resolves `libdd-common@4.2.0`, can you bump it, please?
```suggestion
libdd-common = { version = "4.2.0", default-features = false }
```

iunanua (approved):

datadog-opentelemetry/src/core/configuration/configuration.rs
bm1549: That's correct - updated the docs

PR #232, opened 2026-05-19
chore(msrv)!: update to rust 1.87 [APMSP-3034]

bantonsson:
# What does this PR do?

Bump MSRV from 1.84.1 to 1.87.0.

# Motivation

Upstream dependencies in `libdatadog` now require at least `edition2024`

# Additional Notes

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -46,7 +46,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rust_version: ["1.84.1", "stable", "nightly-2026-04-07"]
+        rust_version: ["1.87.0", "stable", "nightly-2026-04-07"]
         platform: [windows-latest, ubuntu-latest]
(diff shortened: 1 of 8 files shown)


ekump (approved):

.github/workflows/lint.yaml
```diff
@@ -46,7 +46,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rust_version: ["1.84.1", "stable", "nightly-2026-04-07"]
```
iunanua, line 49: could we get the rust_version from Cargo.toml to use it in all the different workflows? does it worth?

iunanua (approved):

.github/workflows/lint.yaml
bantonsson: Maybe we could. But I think we should rethink the interaction with system tests as well in that case.

PR #236, opened 2026-05-26
chore: bump libdd-data-pipeline 3.0.0 -> 5.0.0

Eldolfin:
# What does this PR do?
Update `libdd-data-pipeline` to its latest version and fix breakings.

# Motivation
[Another PR](https://github.com/DataDog/dd-trace-rs/pull/225) will follow this one to expose a new parameter for testing.

# Additional Notes
- Had to inline `wait_agent_info_ready()` because it's now async

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ libdd-capabilities-impl = { version = "2.0.0", default-features = false }
 libdd-telemetry = { version = "5.0.0", default-features = false }
 libdd-common = { version = "4.1.0", default-features = false }
 libdd-tinybytes = { version = "1.1.1", default-features = false }
-libdd-library-config = { version = "1.1.0", default-features = false }
+libdd-library-config = { version = "2.0.0", default-features = false }
 libdd-sampling = { version = "1.0.0", default-features = false }
 opentelemetry_sdk = { version = "0.31.0", features = [
     "trace",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,9 +28,9 @@ libdd-data-pipeline = { version = "5.0.0",  features = [
 ], default-features = false }
 libdd-trace-utils = { version = "5.0.0", default-features = false }
 libdd-capabilities-impl = { version = "2.0.0", default-features = false }
-libdd-telemetry = { version = "4.0.0", default-features = false }
-libdd-common = { version = "3.0.1", default-features = false }
-libdd-tinybytes = { version = "1.1.0", default-features = false }
+libdd-telemetry = { version = "5.0.0", default-features = false }
+libdd-common = { version = "4.1.0", default-features = false }
+libdd-tinybytes = { version = "1.1.1", default-features = false }
 libdd-library-config = { version = "1.1.0", default-features = false }
 libdd-sampling = { version = "1.0.0", default-features = false }
 opentelemetry_sdk = { version = "0.31.0", features = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ authors = ["Datadog Inc. <info@datadoghq.com>"]
 
 [workspace.dependencies]
 # Libdatadog dependencies - change to a stable version once we release
-libdd-data-pipeline = { version = "5.0.0",  features = [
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -23,8 +23,11 @@ authors = ["Datadog Inc. <info@datadoghq.com>"]
...
 [workspace.dependencies]
 # Libdatadog dependencies - change to a stable version once we release
-libdd-data-pipeline = { version = "3.0.1", default-features = false }
-libdd-trace-utils = { version = "3.0.0", default-features = false }
+libdd-data-pipeline = { version = "5.0.0",  features = [
+    "telemetry",
+], default-features = false }
+libdd-trace-utils = { version = "5.0.0", default-features = false }
+libdd-capabilities-impl = { version = "2.0.0", default-features = false }
 libdd-telemetry = { version = "4.0.0", default-features = false }
```
bantonsson: Shouldn't these be bumped as well? I think the transitive dependencies have changed.

Eldolfin: Wasn't sure this was needed. I just updated these too

iunanua: It's not necessary for the build, but it helps you avoid a few duplicate dependencies.

Even so, `libdd-library-config@1.1.0` is pulling `libdd-trace-protobuf@2.0.0` resulting in `libdd-trace-protobuf@2.0.0` and `libdd-trace-protobuf@3.0.2` on the dep tree...

@bantonsson, should we release a `libdd-library-config@2.0.0` to remove the duplication?

bantonsson: I'm just so confused about how the publishing and versioning works right now. It feels like we are digging ourselves into a hole. I think that in this case there is no real problem, since if it was, the code wouldn't compile, but what if it was something more involved.

iunanua: The problem is that `libdd` crates evolve and are published independently; there are many of them, and each developer is free to add dependencies as they see fit, so in the dependency tree, the same crate can end up being installed via multiple different paths.

I wonder if we could consider imposing certain restrictions to make the tree more manageable

Eldolfin: This clippy lint could be relevant: [clippy::multiple_crate_versions](https://rust-lang.github.io/rust-clippy/master/index.html?search=multiple_crate_versions#multiple_crate_versions)

But might need a lot of maintenance for the allow list

iunanua: I have published `libdd-library-config@2.0.0`, you can just update its version here (it doesn't involve any changes to the code)

Eldolfin: Thanks, done

datadog-opentelemetry/src/exporter/mod.rs
```diff
@@ -609,9 +610,14 @@ impl<T: Send + 'static> TraceExporterWorker<T> {
         {
             // Wait for the agent info to be fetched to get deterministic output when deciding
             // to drop traces or not
-            self.trace_exporter
-                .wait_agent_info_ready(Duration::from_secs(5))
-                .unwrap();
+            let start = std::time::Instant::now();
+            let timeout = Duration::from_secs(5);
+            while libdd_data_pipeline::agent_info::get_agent_info().is_none() {
+                if std::time::Instant::now().duration_since(start) > timeout {
```
iunanua:
```suggestion
                if start.elapsed() > timeout {
```

Cargo.toml
```diff
@@ -23,12 +23,15 @@ authors = ["Datadog Inc. <info@datadoghq.com>"]
 
 [workspace.dependencies]
 # Libdatadog dependencies - change to a stable version once we release
-libdd-data-pipeline = { version = "3.0.1", default-features = false }
-libdd-trace-utils = { version = "3.0.0", default-features = false }
-libdd-telemetry = { version = "4.0.0", default-features = false }
-libdd-common = { version = "3.0.1", default-features = false }
-libdd-tinybytes = { version = "1.1.0", default-features = false }
-libdd-library-config = { version = "1.1.0", default-features = false }
+libdd-data-pipeline = { version = "5.0.0",  features = [
```
iunanua:
```suggestion
libdd-data-pipeline = { version = "5.0.0", features = [
```

iunanua (approved):
LGTM

PR #237, opened 2026-05-28
fix(configuration): redact configurations we are not interested in

paullegranddc:
# What does this PR do?

# Motivation

# Additional Notes

### DIFF
--- a/datadog-opentelemetry/src/core/configuration/configuration.rs
+++ b/datadog-opentelemetry/src/core/configuration/configuration.rs
@@ -97,6 +97,17 @@ const RC_DEFAULT_POLL_INTERVAL: f64 = 5.0; // 5 seconds is the highest interval 
 const DEFAULT_UNIX_TRACE_AGENT_URL: &str = "/var/run/datadog/apm.socket";
 const DEFAULT_UNIX_DOGSTATSD_AGENT_URL: &str = "/var/run/datadog/dsd.socket";
 
+/// Placeholder used in telemetry payloads when a sensitive configuration value
+/// has been set
+const REDACTED_VALUE: &str = "[REDACTED]";
+
+/// Controls whether a configuration item's non-default value is visible in telemetry.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum ConfigSensitivity {
+    NonSensitive,
+    Sensitive,
+}
+
 /// OTLP protocol types for OTLP export.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 #[non_exhaustive]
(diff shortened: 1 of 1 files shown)


paullegranddc (commented):
Thanks for the PR! The approach is clean and the test coverage is solid. I have three things to address before this is ready to merge.

---

### 🐛 Dead code — `sampling_rules_item` block

`sampling_rules_item` is built and populated but never used — `cisu.update_parsed(default.trace_sampling_rules)` does the same `get_parse` call independently. This means a redundant parse runs on every `Config` construction. The block (and the now-stale comment `// Use the initialized ConfigItem`) should be removed.

```rust
// All of this is dead — can be deleted
let parsed_sampling_rules_config = sources
    .get_parse::<ParsedSamplingRules>(SupportedConfigurations::DD_TRACE_SAMPLING_RULES);
let mut sampling_rules_item = ConfigItemWithOverride::new_rc( … );
if let Some(rules) = parsed_sampling_rules_config.value {
    sampling_rules_item.set_value_source(rules.value, rules.origin);
}
```

> The Rust compiler does not warn here because calling `.set_value_source()` on the variable counts as "using" it, even though the variable is never read back.

---

### ⚠️ Double-redaction for `Calculated` sensitive values

In `ConfigItemWithOverride::get_all_configurations` the calculated value is redacted **before** being passed to `build_configurations_list`:

```rust
let calculated_option = if self.source() == ConfigSourceOrigin::Calculated {
    let raw = override_value.as_ref().unwrap().get_configuration_value();
    Some(if sensitive { REDACTED_VALUE.to_string() } else { raw })  // redacted here …
} else { None };
```

But `build_configurations_list`'s `make_entry` closure then redacts it *again* (since `Calculated != Default`). Harmless in practice, but confusing. The fix is to pass the raw value through and let `make_entry` own all redaction decisions:

```rust
let calculated_option = if self.source() == ConfigSourceOrigin::Calculated {
    // Pass the raw value; build_configurations_list's make_entry closure
    // handles redaction uniformly for all origins.
    Some(override_value.as_ref().unwrap().get_configuration_value())
} else {
    None
};
```

---

### ⚠️ Document why `trace_agent_url` / `dogstatsd_agent_url` are `NonSensitive`

URLs can technically carry credentials (`http://user:secret@host:port`), so future contributors may question the `NonSensitive` label. A brief comment removes the ambiguity:

```rust
trace_agent_url: ConfigItemWithOverride::new_calculated(
    SupportedConfigurations::DD_TRACE_AGENT_URL,
    Cow::Borrowed(""),
    // The agent is assumed to run on the same host / internal infra as the
    // application, without authentication, so the URL is safe to expose in
    // telemetry.
    NonSensitive,
),
// … same comment for dogstatsd_agent_url
```

---

I've pushed a fixup commit to the branch with all three changes applied. All 295 unit tests still pass.

PR #248, opened 2026-06-08
feat(propagation): add support for DD_TRACE_PROPAGATION_BEHAVIOR_EXTRACT parameter

MilanGarnier:
# What does this PR do?

Implements `DD_TRACE_PROPAGATION_BEHAVIOR_EXTRACT` for dd-trace-rs.

Three modes:

- `continue` (default): existing behavior -- incoming trace context is used as the local trace context. Baggage is propagated.
- `restart`: starts a new trace with a fresh trace ID and sampling decision. The incoming context is referenced via a span link with `reason=propagation_behavior_extract`. Baggage is propagated.
- `ignore`: discards the entire incoming trace context. 

# Motivation

RFC: [Trace Context Propagation Extraction Modes](https://docs.google.com/document/d/1htJFBWR4RFpmK1i6wopxkZwlvYu9unkPJnMnTom4Xjo)

# Additional Notes
- Added tests for parsing the new configuration parameter
- Added integration tests for configurations `restart` and `ignore` 
- Edit in SpanProcessor: Reconstruct the full 128-bit trace ID: `link.trace_id` holds the low 64 bits and `link.trace_id_high` the upper 64. Casting the low bits alone would zero-extend and silently drop the high bits (e.g. an incoming `_dd.p.tid`). It matters since the pr introduces restart links potentially referencing 128bit trace ids


bm1549 (approved):
Everything looks good - great work!

datadog-opentelemetry/src/core/configuration/configuration.rs
```diff
@@ -757,6 +757,61 @@ impl ExtraServicesTracker {
...
+    }
+}
+
+impl ConfigurationValueProvider for TracePropagationBehaviorExtract {
+    fn get_configuration_value(&self) -> String {
+        self.to_string()
+    }
+}
+
+impl ConfigurationValueProvider for Option<TracePropagationBehaviorExtract> {
```
iunanua: is this impl needed?

datadog-opentelemetry/src/core/configuration/configuration.rs
```diff
@@ -757,6 +757,61 @@ impl ExtraServicesTracker {
...
+        };
+        write!(f, "{behavior}")
+    }
+}
+
+impl ConfigurationValueProvider for TracePropagationBehaviorExtract {
+    fn get_configuration_value(&self) -> String {
+        self.to_string()
+    }
+}
```
iunanua: Since `TracePropagationBehaviorExtract` implements `Display`, you should use the `impl_config_value_provider!(simple: …)` macro
```suggestion

```

datadog-opentelemetry/src/text_map_propagator.rs
```diff
@@ -169,31 +170,43 @@ impl DatadogPropagator {
...
+            TracePropagationBehaviorExtract::Restart => match extracted {
+                Some((dd_span_context, parent_style)) => cx.with_value(DatadogExtractData {
+                    links: vec![SpanLink::restart(&dd_span_context, parent_style)],
+                    origin: None,
+                    internal_tags: HashMap::new(),
+                    sampling: Sampling::default(),
+                }),
+                None => cx.clone(),
+            },
+            TracePropagationBehaviorExtract::Ignore => return cx.clone(),
```
iunanua: In `Ignore` mode, the `extracted` result is discarded immediately after running every configured extractor. Parhaps `Ignore` branch should not call `self.inner.extract(&extractor)`.

datadog-opentelemetry/src/text_map_propagator.rs
```diff
@@ -169,31 +170,43 @@ impl DatadogPropagator {
...
-                    trace_state,
-                );
-
-                cx.with_remote_span_context(otel_span_context)
-                    .with_value(DatadogExtractData::from_span_context(dd_span_context))
-            })
-            .unwrap_or_else(|| cx.clone());
-
+        let extracted = self.inner.extract(&extractor);
+        let cx = match self.cfg.trace_propagation_behavior_extract() {
```
iunanua: I'm not sure where this match should be. In theory, `TracePropagationBehaviorExtract` is something specific to datadog, so idk if `DatadogCompositePropagator` would be the right place for it. 
But it would require to change the `DatadogCompositePropagator::extract` signature to something like 
```rust
// ExtractResult or another more appropiate name
enum ExtractResult {
  Continue(SpanContext),
  Restart(SpanLink),
  Passthrough,
  Ignore,
}
```
and handle each case here accordingly.
wdyt @bantonsson @paullegranddc?

MilanGarnier: I was thinking about it, an argument in favor would be that the configuration key would be used in the same place as semantically similar keys such as `TracePropagationStyle`.

MilanGarnier: @bantonsson @paullegranddc @iunanua 
I have implemented this suggestion in [18c6b21](https://github.com/DataDog/dd-trace-rs/pull/248/commits/18c6b210ca72e13de6bf2a0203f983f8b88798d2)
I let you decide which version you think fits best

datadog-opentelemetry/src/propagation/context.rs
```diff
@@ -46,6 +46,18 @@ pub struct SpanLink {
 }
 
 impl SpanLink {
+    /// Creates a span link for a restarted context scenario.
+    pub(crate) fn restart(context: &SpanContext, style: TracePropagationStyle) -> Self {
```
iunanua:
```suggestion
    pub(super) fn restart(context: &SpanContext, style: TracePropagationStyle) -> Self {
```

datadog-opentelemetry/src/propagation/mod.rs
```diff
@@ -49,6 +50,9 @@ pub trait PropagationConfig: Send + Sync {
     /// Whether to stop extraction after the first successful propagator.
     fn trace_propagation_extract_first(&self) -> bool;
 
+    /// Propagation behavior whan extracting: continue, restart or ignore
```
iunanua:
```suggestion
    /// Propagation behavior when extracting: continue, restart or ignore
```

datadog-opentelemetry/src/propagation/mod.rs
```diff
@@ -116,18 +132,31 @@ impl<C: PropagationConfig> DatadogCompositePropagator<C> {
...
     }
 
-    /// Extracts trace context from the carrier using the configured extraction styles.
+    /// Extracts trace context from the carrier using the configured extraction styles, along with
+    /// the style used to extract it.
     ///
     /// Returns `None` if no valid trace context is found in the carrier.
-    pub fn extract(&self, carrier: &dyn Extractor) -> Option<SpanContext> {
+    pub fn extract(&self, carrier: &dyn Extractor) -> ExtractResult {
         let contexts = self.extract_available_contexts(carrier);
```
iunanua, line 144: `self.extract_available_contexts(carrier)` is called when behavior is `TracePropagationBehaviorExtract::Ignore`

https://github.com/DataDog/dd-trace-rs/pull/248#discussion_r3426558938

MilanGarnier: @iunanua  Updated in [2a97724](https://github.com/DataDog/dd-trace-rs/pull/248/commits/2a977240d028ae7c8b18e12998577126bbd562ad)

datadog-opentelemetry/src/span_processor.rs
iunanua, line 592: Oops, I think there is a :bug: here (since we added support for baggage propagation).

If neither `tracecontext` nor `datadog` propagation headers but `baggage` are received, there is no remote context. So here we are not calling `baggage_span_tags` and span attributes are not set :S

Let me fix it in a different PR

datadog-opentelemetry/tests/snapshots/opentelemetry_api/test_injection_extraction_extract_behavior_restart.json
ekump, line 1: I think you can delete this file. It doesn't look like it's used anywhere.

datadog-opentelemetry/tests/snapshots/opentelemetry_api/test_injection_extraction_extract_behavior_restart_tracestats.json
ekump, line 1: I think this snapshot can also be deleted.

datadog-opentelemetry/src/propagation/mod.rs
```diff
@@ -116,18 +132,34 @@ impl<C: PropagationConfig> DatadogCompositePropagator<C> {
         }
     }
 
-    /// Extracts trace context from the carrier using the configured extraction styles.
+    /// Extracts trace context from the carrier using the configured extraction styles, along with
+    /// the style used to extract it.
```
ekump: The style is not returned. Is this a stale doc comment?
```suggestion
/// Extracts trace context from the carrier using the configured extraction styles.                                                                                                                                                                                                      
///                                                                                                                                                                                                                                                                                      
/// Returns an [`ExtractResult`] indicating how the caller should treat the carrier.                                                                                                                                                                                                                                                                                                           
```

datadog-opentelemetry/src/span_processor.rs
ekump: I know this wasn't changed as part of this PR but this doc comment is out of date. CC @iunanua 




```suggestion
/// Adds the span links recorded in the parent context's [`DatadogExtractData`] to `span`.                                                                                                                                                                                               
///                                                                                                                                                                                                                                                                                      
/// These links are produced during extraction — `terminated_context` links for conflicting                                                                                                                                                                                              
/// contexts (continue), or the `propagation_behavior_extract` link (restart). The full 128-bit                                                                                                                                                                                          
/// linked trace ID is reconstructed from `trace_id_high` (high 64) + `trace_id` (low 64).                                                                                                                                                                                               
///                                                                                                                                                                                                                                                                                      
/// Sampling decision, origin and propagation tags from `DatadogExtractData` are handled                                                                                                                                                                                                 
/// separately by the sampler, not here.
```

datadog-opentelemetry/src/propagation/mod.rs
```diff
@@ -116,18 +132,34 @@ impl<C: PropagationConfig> DatadogCompositePropagator<C> {
...
     ///
     /// Returns `None` if no valid trace context is found in the carrier.
-    pub fn extract(&self, carrier: &dyn Extractor) -> Option<SpanContext> {
+    pub fn extract(&self, carrier: &dyn Extractor) -> ExtractResult {
```
ekump, line 150: minor: As is, the `unreachable!()` is genuinely unreachable, but this is kind of brittle in case the code changes in the future. Panics are a big deal in the SDK and since this is a public function there is no guarantee the caller will wrap in a `catch_panic!`. `trace_propagation_behavior_extract()` is also called twice in this function. Both of these indicate a mild code smell.

I think this will achieve the same goal, and remove the unreachable arm:

```Rust
/// Runs the configured extractors, returning `None` when no context was found.
fn extracted_contexts(
    &self,
    carrier: &dyn Extractor,
) -> Option<Vec<(SpanContext, TracePropagationStyle)>> {
    let contexts = self.extract_available_contexts(carrier);
    (!contexts.is_empty()).then_some(contexts)
}

/// Extracts trace context from the carrier using the configured extraction styles.
///
/// Returns an [`ExtractResult`] indicating how the caller should treat the carrier.
pub fn extract(&self, carrier: &dyn Extractor) -> ExtractResult {
    match self.config.trace_propagation_behavior_extract() {
        // Ignore short-circuits before running any extractor.
        TracePropagationBehaviorExtract::Ignore => ExtractResult::Ignore,
        TracePropagationBehaviorExtract::Continue => match self.extracted_contexts(carrier) {
            Some(contexts) => ExtractResult::Continue(Self::resolve_contexts(contexts, carrier)),
            None => ExtractResult::Passthrough,
        },
        TracePropagationBehaviorExtract::Restart => match self.extracted_contexts(carrier) {
            Some(contexts) => {
                let style = contexts[0].1;
                let span_context = Self::resolve_contexts(contexts, carrier);
                ExtractResult::Restart(SpanLink::restart(&span_context, style))
            }
            None => ExtractResult::Passthrough,
        },
    }
}
```

ekump (approved):
Some minor comments, but overall LGTM.

datadog-opentelemetry/tests/snapshots/opentelemetry_api/test_injection_extraction_extract_behavior_restart.json
MilanGarnier: This was indeed a stale test that was used before a previous suggestion, thanks !

datadog-opentelemetry/src/propagation/mod.rs
MilanGarnier: resolved in [0caaa1d](https://github.com/DataDog/dd-trace-rs/pull/248/commits/0caaa1d0e7650ee5646b1da868b275163137b3bf)

PR #250, opened 2026-06-09
refactor(rc): migrate remote config client to libdd-remote-config

iunanua:
## Summary

- Replace the in-tree Remote Configuration client (hyper poller, hand-rolled TUF parsing, base64/SHA256 caching, `ConfigState` bookkeeping, custom `ProductRegistry`, path extraction ~2860 lines) with a thin adapter over `libdd-remote-config::SingleChangesFetcher`.
- `ApmTracingConfig` parser and `apply_apm_tracing` business logic are preserved verbatim and registered via `impl RemoteConfigContent for ApmTracingConfig`, so null-vs-missing semantics, `[0.0, 1.0]` rate validation, service_target gating, env/RC rules composition, and default-provenance catch-all all remain unchanged.
- Drops now-unused deps (`base64`, `sha2`, `hyper-util`, `http-body-util`) and forwards `libdd-remote-config/https` through the local `https` feature.


datadog-opentelemetry/src/core/configuration/remote_config.rs
```diff
@@ -241,50 +94,172 @@ struct LibConfig {
...
+                // produce DM "-3" (LOCAL_USER), not "-12" (REMOTE_DYNAMIC).
+                // Omit `provenance`; libdd-sampling deserializes it as
+                // "default" via its serde default, which maps to DM -3.
+                rules.push(serde_json::json!({ "sample_rate": r }));
+            }
+
+            let rules_json = serde_json::to_string(&serde_json::Value::Array(rules))
+                .map_err(|e| anyhow::anyhow!("Failed to serialize sampling rules: {}", e))?;
+
+            config
```
iunanua, line 271: There is some sort of dance (`Value` -> string -> parse -> `SamplingRuleConfig`) because `update_sampling_rules_from_remote` ends up deserialising `rules_json`.
Do you think it's worth changing it?

datadog-opentelemetry/src/core/configuration/remote_config.rs
```diff
@@ -1,186 +1,36 @@
...
-}
-
-#[derive(Debug, Deserialize)]
-struct TargetFile {
-    /// Path of the file
-    path: String,
-    /// Raw content (base64 encoded in responses)
-    raw: String,
-}
+/// HTTP timeout for a single RC fetch. Matches the historical in-tree client.
```
ekump: I don't think a comment referencing the previous implementation adds any value.

```suggestion
/// HTTP timeout for a single RC fetch
```

datadog-opentelemetry/src/core/configuration/remote_config.rs
```diff
@@ -241,50 +94,172 @@ struct LibConfig {
...
-    #[allow(dead_code)] // Part of TUF specification but not used in current implementation
-    version: Option<u64>,
+/// Apply a parsed APM_TRACING config to the tracer's sampling state.
+///
+/// Returns `Err` on malformed payloads (out-of-range rate, non-numeric rate,
+/// non-array rules, or downstream rule-rejection by libdd-sampling) so the
+/// caller can report `ConfigApplyState::Error` to the agent and leave any
+/// prior remote override in place.
+fn apply_apm_tracing(tracing_config: &ApmTracingConfig, config: &Arc<Config>) -> Result<()> {
+    // Defense-in-depth target check: only apply a config whose service_target
```
ekump: Is having the same [comment](https://github.com/DataDog/dd-trace-rs/pull/250/changes?diff=split#diff-dcc372555a993683bd4ec0e01a2f2aa4010eb9532c33649bb140ae4b70cd279cR53-R57) in two places necessary?

datadog-opentelemetry/src/core/configuration/remote_config.rs
```diff
@@ -241,50 +94,172 @@ struct LibConfig {
...
+
+    let lib = &tracing_config.lib_config;
+
+    let any_field_present =
+        lib.tracing_sampling_rules.is_some() || lib.tracing_sampling_rate.is_some();
+
+    // tracing_sampling_rate must be either null (clear) or a JSON number.
+    // Any other present-but-non-numeric value (string, bool, object) is a
+    // malformed payload — reject rather than silently treating it as a
+    // clear, which would wipe an active remote sampling policy.
```
ekump: Is this level of comment detail necessary? The code doesn't look that complex. I think it could be edited down to `tracing_sampling_rate must be either null (clear) or a JSON number. Otherwise return an error`

datadog-opentelemetry/src/core/configuration/remote_config.rs
```diff
@@ -1,186 +1,36 @@
...
+//! Drives [`libdd_remote_config::fetch::SingleChangesFetcher`] from a dedicated
+//! std thread running a single-threaded Tokio runtime, parses delivered files
+//! via a custom [`ApmTracingConfig`] parser (registered through
+//! [`RemoteConfigContent`]), and routes parsed payloads into
+//! [`Config::update_sampling_rules_from_remote`] / [`Config::clear_remote_sampling_rules`].
+
 use crate::core::configuration::Config;
 use crate::core::utils::{ShutdownSignaler, WorkerHandle};
 
 use anyhow::Result;
```
ekump: Are we violating our own [contributing.md](https://github.com/DataDog/dd-trace-rs/blob/main/CONTRIBUTING.md?plain=1#L134) here?

datadog-opentelemetry/src/core/configuration/remote_config.rs
```diff
@@ -241,50 +94,172 @@ struct LibConfig {
...
-    /// Version of the signed targets
-    #[allow(dead_code)] // Part of TUF specification but not used in current implementation
-    version: Option<u64>,
+/// Apply a parsed APM_TRACING config to the tracer's sampling state.
+///
+/// Returns `Err` on malformed payloads (out-of-range rate, non-numeric rate,
+/// non-array rules, or downstream rule-rejection by libdd-sampling) so the
+/// caller can report `ConfigApplyState::Error` to the agent and leave any
+/// prior remote override in place.
+fn apply_apm_tracing(tracing_config: &ApmTracingConfig, config: &Arc<Config>) -> Result<()> {
```
ekump: nit: This function is rather large. Can it be broken up a bit to be more readable?

datadog-opentelemetry/src/core/configuration/remote_config.rs
```diff
@@ -390,835 +367,174 @@ impl RemoteConfigClientWorker {
...
-                }
-            },
-        )?
-        .split("/");
-
-    let (product, config_id) = (
-        components.next()?.to_string(),
-        components.next()?.to_string(),
+fn apply_remove(file: &Arc<StoredApmFile>, config: &Arc<Config>) {
+    use libdd_remote_config::file_change_tracker::FilePath;
```
ekump: nit: Why do we have use statements inside functions? I also noticed it [here](https://github.com/DataDog/dd-trace-rs/pull/250/changes?diff=split#diff-dcc372555a993683bd4ec0e01a2f2aa4010eb9532c33649bb140ae4b70cd279cR301).

ekump (approved):
non-blocking comments, LGTM

datadog-opentelemetry/src/core/configuration/remote_config.rs
ekump: My two cents: It's probably fine. I think you'll only hit this when config changes, which won't happen that often. The benefit is `apply_apm_tracing()` doesn't have to be coupled to `libdd_sampling::SamplingRuleConfig`

datadog-opentelemetry/src/core/configuration/remote_config.rs
iunanua: moved

datadog-opentelemetry/src/core/configuration/remote_config.rs
iunanua: changed

datadog-opentelemetry/src/core/configuration/remote_config.rs
iunanua: removed

PR #254, opened 2026-06-16
feat(propagation): implement B3 multi-header propagator

duncanista:
## Summary

Stacked on top of #253 (the enum variants). Adds `datadog-opentelemetry/src/propagation/b3multi.rs` with extract/inject/keys, wires it into the `TracePropagationStyle::B3Multi` dispatch, and adds the `"b3multi"` case to `FromStr` now that the propagator actually works (deferred from #253 to avoid the extract-first regression).

Semantics mirror dd-trace-py `_B3MultiHeader` so wire-level behavior matches the other Datadog tracers:

- `extract` returns `None` if `x-b3-traceid` is missing, unparseable, or zero. **A present-but-malformed `x-b3-spanid` also rejects the whole context** (matches dd-trace-py raising `ValueError`); a missing or zero span id is accepted as `0`.
- 128-bit trace ids are kept as a single `u128`; **no** `_dd.p.tid` split and **no** `_dd.parent_id` tag — those are tracecontext-specific in this crate.
- Sampling decode: `x-b3-sampled=0 → AUTO_REJECT`, `=1 → AUTO_KEEP`, `x-b3-flags=1 → USER_KEEP` (debug takes precedence over `sampled`).
- `inject` emits 32 hex chars for 128-bit ids, 16 otherwise; `priority > 1` emits `x-b3-flags: 1` and suppresses `x-b3-sampled` (the two should never both be set per the spec).

## Review feedback addressed (Codex)

- **P2: malformed B3 span IDs**: `extract_malformed_span_id_rejects_context` now asserts that `x-b3-traceid: <valid>; x-b3-spanid: nothex` returns `None` instead of silently coercing span_id to 0.
- The new test `extract_zero_span_id_yields_zero` keeps the "explicit zero is OK" path locked in.

## Test plan

- [x] 16 unit tests in `propagation::b3multi::test` covering: missing/malformed/zero trace id, 64- and 128-bit ids, sampling decode (incl. flags overriding sampled), zero/missing/malformed span id, inject 64-/128-bit, inject priority mapping, `Propagator` dispatch routing, and `keys()` shape.
- [x] `cargo test -p datadog-opentelemetry --lib propagation::` — passes, no regressions.
- [x] `cargo test -p datadog-opentelemetry --lib core::configuration::configuration::tests::test_propagation_style_b3` — new `test_propagation_style_b3multi_parsed_from_env` asserts b3multi now parses while b3 is still dropped (added in #255).
- [x] `cargo clippy -p datadog-opentelemetry --all-targets -- -D warnings` — clean.
- [x] `cargo fmt` with the pinned nightly — clean.

## Base branch

Targets `feat/b3-propagation-style-enum` (#253). Will be retargeted to `main` once that lands.

### DIFF
--- a/datadog-opentelemetry/src/core/configuration/configuration.rs
+++ b/datadog-opentelemetry/src/core/configuration/configuration.rs
@@ -821,12 +821,8 @@ impl FromStr for TracePropagationStyle {
             "datadog" => Ok(TracePropagationStyle::Datadog),
             "tracecontext" => Ok(TracePropagationStyle::TraceContext),
             "baggage" => Ok(TracePropagationStyle::Baggage),
-            // `b3multi` and `b3` are added to from_str alongside the b3multi and b3
-            // single-header propagator implementations. Accepting them here while their
-            // extractors are no-ops would regress users who set
-            // DD_TRACE_PROPAGATION_STYLE_EXTRACT=b3,datadog with
-            // DD_TRACE_PROPAGATION_EXTRACT_FIRST=true — the no-op b3 propagator would be
-            // picked first and Datadog headers would never be tried.
+            "b3multi" => Ok(TracePropagationStyle::B3Multi),
+            // `b3` is added to from_str alongside the b3 single-header propagator.
             "none" => Ok(TracePropagationStyle::None),
             _ => Err(format!("Unknown trace propagation style: '{s}'")),
         }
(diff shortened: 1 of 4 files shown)


duncanista:
Superseded by #258 — branch renamed; this PR's head ref was deleted as a side effect. Replacement carries the same commits.

PR #257, opened 2026-06-16
feat(propagation): add B3Multi and B3SingleHeader variants

duncanista:
## Summary

- Add `B3Multi` and `B3SingleHeader` variants to `TracePropagationStyle`.
- Extend `Display` so the variants render as their canonical Datadog tracer names (`b3multi`, `b3`).
- Wire the new variants into the `Propagator` dispatch as no-ops (extract → `None`, inject → no-op, keys → `&[]`) for exhaustive matching.

**`FromStr` parsing is intentionally deferred** to the per-propagator follow-ups so each style only starts parsing once its extract/inject actually works. Accepting `b3` / `b3multi` here while their dispatch is a no-op would regress users with `DD_TRACE_PROPAGATION_STYLE_EXTRACT=b3,datadog` + `DD_TRACE_PROPAGATION_EXTRACT_FIRST=true` — the no-op b3 would be selected first and Datadog headers would never be tried (raised by Codex review on the initial revision).

## Test plan

- [x] `cargo test -p datadog-opentelemetry --lib core::configuration::` — passes.
- [x] `cargo test -p datadog-opentelemetry --lib propagation::` — passes (existing tests unaffected; new variants currently exercise the no-op dispatch arms).
- [x] `cargo clippy -p datadog-opentelemetry --all-targets -- -D warnings` — clean.
- [x] `cargo fmt` with the pinned nightly — clean.
- [x] New test `test_propagation_style_b3_not_yet_parsed_from_env` asserts that env-var parsing silently drops `b3`/`b3multi` for now, leaving working styles intact.
- [ ] Follow-up PRs add `FromStr` cases alongside the real propagators.

## Follow-ups

- #254: `b3multi.rs` propagator + `"b3multi"` parsing.
- #255: `b3.rs` single-header propagator + `"b3"` parsing.
- #256: Fill in the existing `// todo: add b3` markers in `propagation/mod.rs` (cross-style conflict / span-link tests).

### DIFF
--- a/datadog-opentelemetry/src/core/configuration/configuration.rs
+++ b/datadog-opentelemetry/src/core/configuration/configuration.rs
@@ -785,6 +785,10 @@ pub enum TracePropagationStyle {
     TraceContext,
     /// W3C Baggage propagation format using the `baggage` header.
     Baggage,
+    /// B3 multi-header propagation format using `x-b3-*` headers.
+    B3Multi,
+    /// B3 single-header propagation format using the `b3` header.
+    B3SingleHeader,
     /// No propagation - trace context is not propagated.
     None,
 }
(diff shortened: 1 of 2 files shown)


duncanista:
Collapsed into a single PR per reviewer request — superseded by #262 (branch `jordan.gonzalez/propagation/b3-support`), which carries the same four commits.

bantonsson:
Superseded by #262 which combines all four stacked PRs into one.

PR #265, opened 2026-06-23
fix(baggage): add baggage tags to any root span

iunanua:
# What does this PR do?

Add baggage tags to any root span

# Motivation

If a request arrives with a `baggage` header but no trace context, no remote context is extracted and therefore no remote span context is added by DatadogPropagator [here](https://github.com/DataDog/dd-trace-rs/blob/main/datadog-opentelemetry/src/text_map_propagator.rs#L187) (and `baggage` headers are extracted independently)

Later in `DatadogSpanProcessor::on_start`, `baggage` tags are only set if span context is remote [here](https://github.com/DataDog/dd-trace-rs/blob/main/datadog-opentelemetry/src/span_processor.rs#L572)


datadog-opentelemetry/tests/integration_tests/opentelemetry_api.rs
```diff
@@ -332,3 +332,29 @@ async fn test_trace_writer_synchronous_mode() {
     }
     test_agent.assert_snapshot(SESSION_NAME).await;
 }
+
```
ekump, line 337: For baggage tags, as I understand there are three scenarios:
1. Request comes in with baggage header and trace context meaning the new span is a local root.
2. Request comes in with baggage header and no trace context, also meaning the new span is a local root.
3.  An existing local span.

You added a test for 2. Should we also have unit tests for 1 and 3?

datadog-opentelemetry/src/span_processor.rs
```diff
@@ -569,15 +569,17 @@ impl opentelemetry_sdk::trace::SpanProcessor for DatadogSpanProcessor {
...
-            }
         } else if !parent_ctx.has_active_span() {
             self.registry.register_local_root_span(trace_id, span_id);
         } else {
             self.registry
                 .register_span(trace_id, span_id, EMPTY_PROPAGATION_DATA);
+            return;
+        }
+
+        // Apply baggage tags to any root span
```
ekump, line 599: Minor: The early return works fine today, and if we have sufficient tests we're probably ok. But, if this function grows (plausible since it's `on_start`), this early return could be a footgun? Would something like this be better:

```Rust
let is_local_root = if parent_ctx.span().span_context().is_remote() {                                                                                                                                                                                                                        
    self.add_remote_links(span, parent_ctx);                                                                                                                                                                                                                                                 
    self.registry.register_local_root_span(trace_id, span_id);                                                                                                                                                                                                                               
    true                                                                                                                                                                                                                                                                                     
} else if !parent_ctx.has_active_span() {                                                                                                                                                                                                                                                    
    self.registry.register_local_root_span(trace_id, span_id);                                                                                                                                                                                                                               
    true                                                                                                                                                                                                                                                                                     
} else {                                                                                                                                                                                                                                                                                     
    self.registry.register_span(trace_id, span_id, EMPTY_PROPAGATION_DATA);                                                                                                                                                                                                                  
    false                                                                                                                                                                                                                                                                                    
};                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                               
  // Baggage tags go on local root spans only                                                                                                                                                                                                                                                  
  if is_local_root {                                                                                                                                                                                                                                                                           
      for kv in baggage_span_tags(parent_ctx.baggage(), self.config.trace_baggage_tag_keys()) {                                                                                                                                                                                                
          span.set_attribute(kv);                                                                                                                                                                                                                                                              
      }                                                                                                                                                                                                                                                                                        
  }
```

datadog-opentelemetry/tests/integration_tests/opentelemetry_api.rs
ekump: Also, out of scope for this PR but should there be a system test for this behavior?

ekump (approved):
a few minor comments, but LGTM.

datadog-opentelemetry/tests/integration_tests/opentelemetry_api.rs
iunanua: Your comment made me check the system tests and they are still disabled (https://github.com/DataDog/system-tests/pull/7207) and there is no test case for 2

Then I modified `test_baggage_span_tags_all` removing `x-datadog-*` headers and ran it with nodejs and java and they failed sooo It makes me think that maybe it's not a bug after all but a feature...

@rachelyangdog can you clarify whether this PR is necessary or not?

rachelyangdog: Hi, ~~I think this PR is correct.~~

~~Baggage tags should be applied to any local root span, regardless of whether upstream trace context is present. Per the RFC: "Whenever a request is received, parse its baggage header to extract key-value pairs" and that tagging happens "at the entry point of a service." Scenario 2 (baggage + no trace context) is still a service entry point so the span being tagged is the one this service creates, not an upstream span.~~

~~I looked at the go and python tracers and both implementations do this. As for java and node, this must be a bug. Thanks for catching it and I can make the changes in those repos. I'll also add a system-test for this case as well.~~

In terms of unit tests for 1 and 3, I agree that we should also have them. Thanks again!!

iunanua: Perfect! Thank you for confirming that.

I have added the tests

iunanua: Oh, I just refreshed the page and I see that you crossed out part of your comment 
What does it mean? is the PR correct or not? :sweat_smile:

rachelyangdog: Sorry, I realized that the RFC I wrote was very blurry on this part of the spec. My comment holds and I think this PR is correct. I wrote a parametric system-test for this scenario and every tracer failed, but I realized it's because of the parametric server and not the tracer implementations. 

Sorry for the confusion! LGTM. Thank you!!

iunanua: then :ship:

PR #266, opened 2026-06-23
feat(config): support OTEL_SERVICE_NAME as alias for DD_SERVICE

bm1549:
# What does this PR do?

Registers `OTEL_SERVICE_NAME` as a first-class configuration entry, with `DD_SERVICE` taking precedence and `OTEL_SERVICE_NAME` used as a fallback when `DD_SERVICE` is absent or empty. This matches how other Datadog tracers (Python, Go, Java, .NET, Node.js) handle it.

Previously, `OTEL_SERVICE_NAME` was not a tracer config value. It was picked up only via the OpenTelemetry SDK's resource detection and applied post-init in `set_resource`, as a `Calculated`-origin override on the service config plus the span resource. Because the Remote Config worker starts before `set_resource` runs and builds its request lazily, the initial RC request(s) could be sent with the wrong (default or `DD_SERVICE`) service name. Resolving `OTEL_SERVICE_NAME` at config-initialization time removes that startup window, so Remote Config targeting and all config-level consumers use the correct service name from the start.

**Changes:**
- `supported-configurations.json`: Adds `OTEL_SERVICE_NAME` as a top-level entry (version B, matching the FPD registry), alongside the existing OTel config entries. It is registered as a separate config rather than a `DD_SERVICE` alias because the FPD treats OTel env vars as distinct configs, not aliases (using the `aliases` field caused the FPD validation CI job to fail).
- `configuration.rs`: Adds a `update_non_empty_string_with_fallback` helper on `ConfigItemSourceUpdater` and uses it for the `service` field. The service is the first non-empty value among `[DD_SERVICE, OTEL_SERVICE_NAME]`, else the default. An empty value on either key is treated as unset.
- `supported_configurations.rs`: Regenerated, `OTEL_SERVICE_NAME` is now a `SupportedConfigurations` enum variant.

# Motivation

Users who configure their services via `OTEL_SERVICE_NAME` (a standard OpenTelemetry env var) expect dd-trace-rs to honor it the same way the other Datadog tracers do.

# Additional Notes

Service-name precedence tests cover:
1. `OTEL_SERVICE_NAME` alone is used as the service name
2. `DD_SERVICE` + `OTEL_SERVICE_NAME` means `DD_SERVICE` wins
3. `DD_SERVICE=""` + `OTEL_SERVICE_NAME` set falls through to `OTEL_SERVICE_NAME` (empty `DD_SERVICE` is treated as unset, matching dd-trace-go)
4. `OTEL_SERVICE_NAME=""` alone uses the default
5. Both empty uses the default

### DIFF
--- a/datadog-opentelemetry/src/core/configuration/configuration.rs
+++ b/datadog-opentelemetry/src/core/configuration/configuration.rs
@@ -2829,7 +2829,7 @@ mod tests {
 
     #[test]
     fn test_otel_service_name_alias() {
-        // OTEL_SERVICE_NAME is a non-deprecated alias for DD_SERVICE
+        // OTEL_SERVICE_NAME is an alias for DD_SERVICE
         let mut sources = CompositeSource::new();
         sources.add_source(HashMapSource::from_iter(
             [("OTEL_SERVICE_NAME", "otel-service")],
--- a/datadog-opentelemetry/src/core/configuration/configuration.rs
+++ b/datadog-opentelemetry/src/core/configuration/configuration.rs
@@ -622,6 +622,36 @@ impl ConfigItemSourceUpdater<'_> {
         }
     }
 
+    /// Like [`Self::update_non_empty_string`], but consults a `fallback` key when the
+    /// ConfigItem's own key is entirely absent. The primary key takes precedence; an
+    /// explicit empty value on the primary key produces the default (it does not fall
+    /// through to the fallback). The fallback is also required to be non-empty.
+    pub fn update_non_empty_string_with_fallback<ParsedConfig, ConfigItemType, F>(
+        &self,
+        default: ConfigItemType,
+        fallback: SupportedConfigurations,
+        transform: F,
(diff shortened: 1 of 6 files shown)


datadog-opentelemetry/src/core/configuration/configuration.rs
```diff
@@ -622,6 +622,35 @@ impl ConfigItemSourceUpdater<'_> {
         }
     }
 
+    /// Like [`Self::update_non_empty_string`], but uses the first non-empty value among the
+    /// ConfigItem's own key and a `fallback` key (the primary takes precedence). An empty
+    /// value on either key is treated the same as the key being absent. If neither key has a
+    /// non-empty value, the default is returned.
+    pub fn update_non_empty_string_with_fallback<ParsedConfig, ConfigItemType, F>(
```
bm1549, line 629: Fallback behavior is a not fully consistent across languages, but this felt the most obvious to me

bantonsson (approved):

PR #269, opened 2026-06-26
feat: otel thread-level context sharing

yannham:
# What does this PR do?

This PR publishes the current opentelemetry-rust context as an [OTel thread-level context](https://github.com/open-telemetry/opentelemetry-specification/pull/4947) as well. The changes rely for now of yet unmerged changes in opentelemetry-rust and libdatadog.

Since the patch to opentelemetry-rust was developed from a recent version, there's also a bit of unrelated boilerplate changes corresponding to upgrading to the latest opentelemetry-rust.

# Motivation

Adding support for the thread-context level OTEP will enhance the eBPF full host profiler with contexts.

# Additional Notes

### DIFF
--- a/datadog-opentelemetry/src/thread_ctx.rs
+++ b/datadog-opentelemetry/src/thread_ctx.rs
@@ -1,0 +1,139 @@
+// Copyright 2025-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+//! Per-thread OTel context publishing for the eBPF profiler.
+//!
+//! This module keeps the per-thread `otel_thread_ctx_v1` TLS slot (defined in [OTEP #4947]) in sync
+//! with the OTel span that is currently active on each thread. An out-of-process reader such as the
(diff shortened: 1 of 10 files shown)


datadog-opentelemetry/src/thread_ctx.rs
```diff
@@ -0,0 +1,139 @@
...
+/// `observer_cx_view` slot on first use. Returns `None` when `cx` has no valid span — in that case
+/// the caller should detach the TLS slot.
+fn view_thread_ctx(cx: &Context, registry: &TraceRegistry) -> Option<Arc<ThreadContext>> {
+    let span = cx.span();
+    let span_ctx = span.span_context();
+    if !span_ctx.is_valid() {
+        return None;
+    }
+    let trace_id = span_ctx.trace_id().to_bytes();
+    let span_id = span_ctx.span_id().to_bytes();
```
bantonsson, line 72: Can these be moved into the `init` part? They feel unnecessary if the view is already initialized.

PR #271, opened 2026-06-30
feat(config): support OTEL_PROPAGATORS as propagation style fallback

bm1549:
# What does this PR do?

Adds support for the OpenTelemetry `OTEL_PROPAGATORS` env var as a fallback for `DD_TRACE_PROPAGATION_STYLE`. When `DD_TRACE_PROPAGATION_STYLE` is unset, dd-trace-rs reads `OTEL_PROPAGATORS` and applies it to the global propagation style, matching the other Datadog tracers (Python, Go, Java, Node.js, .NET, Ruby, PHP).

It reuses the existing `update_non_empty_string_with_fallback` helper (the one behind the `OTEL_SERVICE_NAME` alias), so only the global style is wired; extract/inject inherit it through the existing fallback in `propagation/config.rs`. Supported values: `datadog`, `tracecontext`, `baggage`, `b3` (single-header), `b3multi`, `none`. Unsupported OpenTelemetry propagators (`jaeger`, `xray`, `ottrace`) are dropped with a warning. `DD_TRACE_PROPAGATION_STYLE` wins when both are set.

# Motivation

Users who configure propagation via `OTEL_PROPAGATORS` expect dd-trace-rs to honor it the way the other tracers do. It was the only tracer (besides dd-trace-cpp) missing it.

# Additional Notes

- An empty `DD_TRACE_PROPAGATION_STYLE=""` is now treated as unset (falls back to `OTEL_PROPAGATORS`, then the default), consistent with the `OTEL_SERVICE_NAME` fallback.
- Companion system-tests PR: DataDog/system-tests#7241.


datadog-opentelemetry/src/core/configuration/configuration.rs
```diff
@@ -1353,10 +1353,18 @@ impl Config {
...
-            trace_propagation_style: cisu
-                .update_parsed_with_transform(default.trace_propagation_style, |DdTags(tags)| {
-                    TracePropagationStyle::from_tags(Some(tags))
-                }),
+            // OTEL_PROPAGATORS is an alias/fallback for DD_TRACE_PROPAGATION_STYLE. An empty
+            // value on either key is treated as unset (consistent with the OTEL_SERVICE_NAME
+            // fallback), so the comma-split is done in the transform rather than via DdTags.
+            trace_propagation_style: cisu.update_non_empty_string_with_fallback(
+                default.trace_propagation_style,
+                SupportedConfigurations::OTEL_PROPAGATORS,
```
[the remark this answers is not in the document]
bm1549: Checked the other tracers on this. For this exact case (`DD_TRACE_PROPAGATION_STYLE_EXTRACT` set, global + `_INJECT` unset, `OTEL_PROPAGATORS` set), inject also inherits `OTEL_PROPAGATORS` in `dd-trace-py`, `dd-trace-go`, `dd-trace-java`, and `dd-trace-js` — only php/rb fall back to the default. Here `OTEL_PROPAGATORS` is a fallback for the global `DD_TRACE_PROPAGATION_STYLE`, and extract/inject inherit the global when their own key isn't set (existing `get_extractors`/`get_injectors` behavior), so this matches the reference impls. Agreed the php/rb split isn't ideal and probably deserves a cross-tracer alignment, but that's out of scope for this PR — keeping as-is here.

datadog-opentelemetry/src/core/configuration/configuration.rs
```diff
@@ -1353,10 +1353,18 @@ impl Config {
...
+                |styles| {
+                    TracePropagationStyle::from_tags(Some(
+                        styles.split(',').map(|s| s.to_string()).collect(),
```
iunanua: it's a bit ugly but continues using the `DdTags`...
```suggestion
                        DdTags::from_str(&styles).unwrap().0,
```

datadog-opentelemetry/src/core/configuration/configuration.rs
```diff
@@ -4345,4 +4394,132 @@ mod tests {
...
+            Some(vec![TracePropagationStyle::Datadog]).as_deref()
+        );
+    }
+
+    #[test]
+    fn test_otel_propagators_filters_unsupported_values() {
+        // Unknown propagator names (e.g. "jaeger") are dropped; supported ones are kept.
+        let mut sources = CompositeSource::new();
+        sources.add_source(HashMapSource::from_iter(
+            [("OTEL_PROPAGATORS", "jaeger,tracecontext")],
```
iunanua, line 4449: when `DD_TRACE_PROPAGATION_STYLE` is unset and `OTEL_PROPAGATORS=unsupported_style`, I think `config.trace_propagation_style()` returns an empty array. Should it return the default values instead?

iunanua (approved):
Is possible to enable the corresponding system tests? Although I've taken a quick look, it seems to me that to enable them, we need to support other otel env vars.

datadog-opentelemetry/src/core/configuration/configuration.rs
bm1549: Done in 9667a09.

datadog-opentelemetry/src/core/configuration/configuration.rs
bm1549: yep good point, done in 9667a09 - all-unsupported `OTEL_PROPAGATORS` now falls back to the default instead of an empty list.

bm1549:
@iunanua Dug into this - it's mostly harness gaps, not missing OTEL vars. The `test_otel_env_vars.py` cases assert on `dd_trace_sample_rate` (the rust parametric `/trace/config` returns `null`), and the `test_telemetry.py` cases need the `otel.env.hiding`/`otel.env.invalid` metrics that aren't wired up here - so enabling them is a follow-up, out of scope for this PR.

PR #274, opened 2026-07-07
refactor(propagation): mute telemetry errors with empty trace_id

iunanua:
# What does this PR do?

- Change `datadog` propagator to check if `trace_id` is empty before triggering an Error
- Help debug by adding the trace_id value in a log

### DIFF
--- a/datadog-opentelemetry/src/propagation/datadog.rs
+++ b/datadog-opentelemetry/src/propagation/datadog.rs
@@ -263,8 +263,10 @@ fn extract_trace_id(carrier: &dyn Extractor) -> Result<Option<u64>, Error> {
     }
 
     let trace_id = trace_id.parse::<u64>().map_err(|_| {
-        dd_warn!("Propagator (datadog): Failed to decode `trace_id` {trace_id}");
-        Error::extract("Failed to decode `trace_id`", "datadog")
+        Error::extract(
+            format!("Failed to decode `trace_id`: \"{trace_id}\""),
+            "datadog",
+        )
     })?;
     if trace_id == 0 {
         return Err(Error::extract("Invalid `trace_id` found", "datadog"));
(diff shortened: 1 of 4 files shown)


datadog-opentelemetry/src/propagation/datadog.rs
```diff
@@ -257,9 +257,15 @@ fn extract_trace_id(carrier: &dyn Extractor) -> Result<Option<u64>, Error> {
...
-        .map_err(|_| Error::extract("Failed to decode `trace_id`", "datadog"))?;
+    if trace_id.is_empty() {
+        dd_warn!("Propagator (datadog): Empty `trace_id` found in datadog carrier");
+        return Ok(None);
+    }
+
+    let trace_id = trace_id.parse::<u64>().map_err(|_| {
+        dd_warn!("Propagator (datadog): Failed to decode `trace_id` {trace_id}");
+        Error::extract("Failed to decode `trace_id`", "datadog")
+    })?;
```
paullegranddc, line 270: This is going to log twice for the same error, once here with the trace_id and once in `extract` with `Error extracting trace_id` 🤔

I think i'd prefer if we either  
* put the faulty trace id value in the error so it is logged at the top level
* returned `Ok(None)` here in order to not double log

iunanua: yes, but this `dd_warn!` prints the trace_id malformed value and the top level prints just the template without the value that I'm interested in.

If we put the trace_id value in the error, we are going to send the value via telemetry and we should send just templates.
If we return `Ok(None)`, the error won't be sent via telemetry

iunanua: Should we add a flag in the `Error` to avoid logging it? and then something like `Error::extract_muted()`

iunanua: umm, maybe it's not that easy...

paullegranddc: > If we return Ok(None), the error won't be sent via telemetry

We will loose the `dd_warn!` in `extract`, but the one in `extrace_trace_id` that you added should have all the info we need

paullegranddc: >Should we add a flag in the Error to avoid logging it? and then something like Error::extract_muted()

I think the long term solution would be to have two parts in errors, one we log only for customers and send to telemetry as a template, and a public one we can include in the telemetry log.

But let's not solve that in this PR

iunanua: I have removed the `dd_warn!` and included the trace_id in the inner Error like this
`Error::extract(format!("Failed to decode `trace_id`: \"{trace_id}\""), "datadog")`

The thing is that error is not expanded when `dd_error!` is called so it is safe to include dynamic messages in `Error`.

In the end, the template sent via telemetry is the same as before when the error reaches
`dd_error!("Propagator (datadog): Error extracting trace_id {e}");`, and the printed log contains the malformed `trace_id` value.

wdyt @paullegranddc?

PR #275, opened 2026-07-09
refactor(rc): run remote config on the trace exporter's shared runtime

iunanua:
## Summary

Follows #250. Both the **Remote Config client** and the **telemetry-metrics collector** used to run on their own dedicated threads — RC on a std thread with its own current-thread tokio runtime, telemetry on a std thread with a `recv_timeout` loop. This moves both onto the single multi-thread runtime already owned by the trace exporter, so all three background loops share one runtime and one teardown path.

### What changed

- The shared `BasicRuntime` is built inside the exporter's dedicated `datadog-trace-init` thread (`build_on_dedicated_thread`), so it never exists on the caller's thread. `DatadogExporter` owns it and drops it off-thread, which keeps a tokio-runtime drop off the caller on every init path: a build failure drops it on the dedicated thread, a thread-spawn failure never creates it, and on success `Drop for DatadogExporter` drops it off-thread. (Dropping a tokio runtime inline runs a blocking shutdown that panics when construction happens inside an existing runtime — which the public builder's `catch_unwind` would turn into a fully no-op provider, discarding the user's provider.)
- `DatadogSpanProcessor::new` builds the exporter first, then registers the **Remote Config** and **telemetry-metrics** [`Worker`]s on the exporter's runtime handle (`span_exporter.shared_runtime()`), mirroring the trace-export worker. Each returned `WorkerHandle` is discarded; teardown happens through the runtime's `shutdown_async` (driven by the exporter's shutdown), so both subsystems drop all of their former cancel-token / channel / thread-join shutdown wiring.
- Each worker's poll/emit interval is built lazily on the first `Worker::trigger`, because tokio timers must be constructed inside a runtime context and `start` runs on the caller's (non-runtime) thread. First-tick timing is preserved to match the old loops: RC fires an immediate startup fetch, and telemetry's first emit is one `EMIT_INTERVAL` (10s) out.
- `trigger` is the cancellation point; a fetch/emit already in `run()` completes before the worker is paused. RC's fetch is bounded by the 3s endpoint timeout, so shutdown can't hang on it.
- Runtime is sized `worker_threads(2)` — scheduler headroom for the combined task mix (trace-export worker + RC + telemetry + the exporter's own internal tasks), which are almost always parked on a timer. Note this is pool capacity, not a thread per worker: tokio's multi-thread scheduler has no worker→thread affinity.
- Deletes `core/utils.rs` (`WorkerHandle` / `ShutdownSignaler` / `WorkerError`), now unused, and updates the `remote_config` module docs to describe the shared-runtime `Worker` model.
- Adds `async-trait` as a direct dep (needed to `impl` the `#[async_trait]` `Worker`; already present transitively, so `LICENSE-3rdparty.csv` is unchanged).

Net **−168 lines**.


iunanua:
@codex review

datadog-opentelemetry/Cargo.toml
bantonsson: `tokio-util` doesn't seem to be used...

bantonsson (approved):

datadog-opentelemetry/Cargo.toml
iunanua: one down!

PR #276, opened 2026-07-09
feat(telemetry): shutdown with a deadline

iunanua:
# What does this PR do?

This wires telemetry shutdown into `DatadogSpanProcessor` shutdown, bounded by the caller's deadline.

Multiple span processors share one process-global telemetry worker; a type-safe `TelemetryUser` token (handed out by register, consumed by trigger_stop) makes the count underflow-safe by construction, so the worker stops only when the last processor shuts down.


paullegranddc (commented):
What happens if I create a tracer, shut it down then create a new one latter?
Will the telemetry worker restart, and still allow sending telemetry

iunanua:
> What happens if I create a tracer, shut it down then create a new one latter? Will the telemetry worker restart, and still allow sending telemetry

Nope, that's one suggestion made by codex. But I ruled it out because it's an unlikely scenario. Is it worth it?

paullegranddc:
> I ruled it out because it's an unlikely scenario. Is it worth it?

I'm not sure, but you could imagine a situation where the tracer is not used as a global, but a scoped object.

We can tolerate not receiving telemetry if a user restarts a tracer, but I think we should at least make sure that using a shutdown telemetry handle does not panic or cause memory leaks.
Also all the test you added test shut down against the `TestHandle`, not against the actual telemetry runtime. I am not sure that enough.
I'd like to see integration tests

iunanua:
> > I ruled it out because it's an unlikely scenario. Is it worth it?
> 
> I'm not sure, but you could imagine a situation where the tracer is not used as a global, but a scoped object.

Tracers (both global and scope-specific) are registered as telemetry users, so telemetry keeps track of active users and only shuts down when the last active user requests a shutdown.

> We can tolerate not receiving telemetry if a user restarts a tracer, but I think we should at least make sure that using a shutdown telemetry handle does not panic or cause memory leaks. Also all the test you added test shut down against the `TestHandle`, not against the actual telemetry runtime. I am not sure that enough. I'd like to see integration tests

I have added integration test to prove that:
- different combinations of scoped and global tracers shutdown correctly
- calling `shutdown` twice or multiple times (from different tracers) does not panic
- dropping a scoped tracer calls `shutdown` and there is no problem when dropping again another scoped tracer
- shutdown has no problems with telemetry off
- concurrent shutdown has no problems
- calling `shutdown` twice from the same tracer does not panic

@paullegranddc let me know if you can think of any other interesting tests.

datadog-opentelemetry/src/core/telemetry.rs
```diff
@@ -20,9 +23,26 @@ use super::telemetry_session;
 use crate::{dd_debug, dd_error, dd_warn};
 
 static TELEMETRY: TelemetryCell = OnceLock::new();
+static TELEMETRY_USERS: AtomicUsize = AtomicUsize::new(0);
 
 type TelemetryCell = OnceLock<Mutex<Telemetry>>;
 
+#[must_use = "dropping a TelemetryUser without calling `trigger_stop` leaks the registration and prevents telemetry from ever shutting down"]
```
bantonsson: This doesn't really catch `trigger_stop` being called. The way `TelemetryUser` is being used, wouldn't it make more sense to implement `Drop` for it?

bantonsson (commented):
Overall good. Only one issue.

datadog-opentelemetry/src/core/telemetry.rs
iunanua: `Drop` implemented

bantonsson (approved):

PR #279, opened 2026-07-17
ci: release automation

iunanua:
# What does this PR do?

- Remove hardcoded version from integration test files snapshots
- Add the `prepare-release.sh` script to automate manual tasks needed to prepare a new release: 
  - Verify rustdocs
  - Replace versions found in README.md and other files
  - Bump Cargo.toml versions
  - Update CHANGELOG entries
- Add cargo release and git cliff config files

PR #280, opened 2026-07-27
ci(release): add release-proposal-dispatch workflow

iunanua:
## What

Adds the **Release proposal** GitHub workflow that drives `scripts/prepare-release.sh` (merged in #279), plus its supporting pieces:

- `.github/workflows/release-proposal-dispatch.yaml` — `workflow_dispatch` → resolve toolchain → install `cargo-release`/`git-cliff` → `prepare-release.sh` → signed commit via `commit-headless` → PR via `octo-sts`.
- `.github/workflows/chainguard/self.write.pr.sts.yaml` — the `self.write.pr` STS policy used to mint the PR token.
- `prepare-release.sh` — new `--no-codex-check` flag (the workflow passes it, since the PR checklist carries the Codex review).

## How to use

- **Normal release:** dispatch the workflow **from `main`** with a `version` (`minor`, `patch`, `major`, or an exact `0.6.0` / `0.6.0-rc.1`). It opens a `chore(release): proposal vX.Y.Z` PR.
- **Hotfix:** still dispatch **from `main`**, but set the `base-branch` input to the hotfix line (e.g. `hotfix/0.5.x`). The STS policy requires a protected branch, so the workflow must always be *dispatched* from `main`; `base-branch` controls what gets checked out and where the PR is targeted.

## Testing from a non-protected branch

No STS token can be minted off a protected branch, so the workflow **falls back to a test run**: it pushes the proposal branch **unsigned** and **skips PR creation**, logging a `::warning::`/`::notice::` explaining what was skipped. This lets the full prepare → commit → push flow be exercised without signing/PR permissions.

## Notes

- Does **not** tag or publish — after the proposal PR merges, tag the commit to trigger `publish.yaml`.
- Stacked on the release-automation work merged in #279.

### DIFF
--- a/.github/workflows/release-proposal-dispatch.yaml
+++ b/.github/workflows/release-proposal-dispatch.yaml
@@ -1,0 +1,166 @@
+name: Release - Open a release PR
+
+# Opens a `chore(release): proposal vX.Y.Z` pull request that bumps the crate
+# version and updates all version references (docs + changelog). It does NOT
+# tag or publish: after this PR is merged, tag the commit to trigger publishing
+# (see docs/releasing.md and .github/workflows/publish.yaml).
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Semver level (major | minor | patch | rc | beta | alpha) or an exact version (e.g. 0.6.0 or 0.6.0-rc.1)"
+        required: true
+        type: string
+      base-branch:
+        description: "Branch to release from and target the PR at. Use a hotfix/release branch to cut a hotfix from an older line."
+        required: false
+        type: string
+        default: main
+
+jobs:
+  propose:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write # Enable OIDC
+      contents: write
+      pull-requests: write
+    env:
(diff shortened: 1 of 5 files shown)


iunanua:
@codex review

hoolioh (approved):

.github/workflows/release-proposal-dispatch.yaml
```diff
@@ -0,0 +1,181 @@
...
+  group: release-proposal-dispatch-group
+  cancel-in-progress: false
+
+jobs:
+  propose:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write # Enable OIDC
+      contents: write
+      pull-requests: write
```
hoolioh: Is write needed? Aren't all modifications carried out by the token?

hoolioh (approved):
LGTM, just minor comment

.github/workflows/release-proposal-dispatch.yaml
iunanua: You are right, it is not needed.
thkx!

PR #281, opened 2026-07-30
feat(log): route the tracer's own diagnostics through tracing

iunanua:
## What

`dd_debug!` / `dd_info!` / `dd_warn!` / `dd_error!` wrote directly to stdout and stderr, which applications could not capture, format, filter, or redirect. They now emit through `tracing` under the `datadog_opentelemetry` target when a subscriber is available.

`DD_LOG_LEVEL` still decides which diagnostics are created. Once created, the application's subscriber controls their formatting, destination, and filtering. For example, debug diagnostics require both `DD_LOG_LEVEL=debug` and a subscriber filter that admits `datadog_opentelemetry=debug`.

Applications built around the `log` facade can enable the new opt-in `log-compat` feature. When `tracing` cannot carry a diagnostic, the feature routes it through `log`; without the feature, the existing stdout/stderr behavior remains as the fallback.

## Routing

For each diagnostic enabled by `DD_LOG_LEVEL`:

1. Error templates are forwarded to instrumentation telemetry as before. This is independent of human-readable log routing, and only the uninterpolated template is sent.
2. If the calling thread has a `tracing` subscriber and the event is within `tracing`'s compile-time level ceiling, emit it through `tracing`.
3. Otherwise, emit through the `log` facade when `log-compat` is enabled, or fall back to stdout/stderr when it is not.

The fallback is used only when `tracing` is incapable of carrying the event:

- **No subscriber on the calling thread.** `dispatcher::has_been_set()` cannot answer this because it remains true after a scoped `with_default` exits and on threads that never had that subscriber, including the tracer's workers.
- **A compile-time ceiling below the event's level.** Additive features such as `tracing/release_max_level_warn` can compile callsites out; no runtime filter can restore them.

If an installed subscriber filters this target or level out, that decision is respected and no fallback is emitted.

OpenTelemetry suppression does **not** prevent dispatch to `tracing`. Non-OpenTelemetry layers can still capture the diagnostic, while `SdkLogger::event_enabled` independently prevents an OpenTelemetry log bridge from exporting it and creating a telemetry loop.

## Decisions

- **`log-compat` is opt-in.** It gives applications using the `log` facade ownership of the fallback without changing default behavior for applications that have no logger installed.
- **`tracing`'s `log` feature is not enabled.** Cargo feature unification could otherwise enable `log` emission for every tracing-instrumented crate in the application's graph and cause duplicate records.
- **`LOG_TARGET = env!("CARGO_CRATE_NAME")`.** This uses the underscored crate name without a duplicated literal. `module_path!()` would identify `core::log`, rather than the code that produced the diagnostic, so the original file and line are attached as event fields.
- **Telemetry forwarding remains unconditional** and happens independently of the human-readable destination.
- **Trace-export errors log inside an OpenTelemetry-suppressed scope.** The SDK does this around its own exporters, but Datadog span export runs on libdatadog's worker. The guard is `!Send`, so it wraps synchronous logging rather than the asynchronous export.

### DIFF
--- a/datadog-opentelemetry/examples/propagator/src/server.rs
+++ b/datadog-opentelemetry/examples/propagator/src/server.rs
@@ -27,7 +27,9 @@ use opentelemetry_stdout::{LogExporter, SpanExporter};
 use std::{convert::Infallible, net::SocketAddr, sync::OnceLock};
 use tokio::net::TcpListener;
 use tracing::info;
-use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
+use tracing_subscriber::{
+    layer::SubscriberExt, util::SubscriberInitExt, EnvFilter, Layer as TracingLayer,
+};
 
 fn get_tracer() -> &'static BoxedTracer {
     static TRACER: OnceLock<BoxedTracer> = OnceLock::new();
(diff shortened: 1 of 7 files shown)


iunanua:
@codex review

iunanua:
@codex review

hoolioh (approved):

datadog-opentelemetry/examples/propagator/src/server.rs
```diff
@@ -242,8 +244,41 @@ fn init_logs() -> SdkLoggerProvider {
...
+
+    // The tracer emits its own diagnostics through `tracing`, under the `datadog_opentelemetry`
+    // target, so this subscriber decides where they end up. How verbose they are is a separate
+    // question, answered by `DD_LOG_LEVEL` / `set_log_level_filter` — see `init_tracer`.
+    //
+    // This bridge turns events into OpenTelemetry log records, which then have to be exported. The
+    // crates the exporters themselves are built on are excluded for that reason: bridging their
+    // events would make every export produce more records to export.
+    // `libdd` covers every `libdd_*` crate, since directives match targets by prefix. Those crates
+    // are the trace transport, and they log from async tasks where the SDK's context-based
```
hoolioh: Comments seem to verbose. Rather than pointing out key points it exposes too much context. It migth be worth reduce it a little bit.

datadog-opentelemetry/src/span_exporter.rs
```diff
@@ -567,6 +568,15 @@ fn log_trace_exporter_error(e: &TraceExporterError) {
...
     use crate::{dd_debug, dd_error};
 
+    // Suppress OpenTelemetry log records for the duration of this call. These diagnostics reach the
+    // application's `tracing` subscriber, which may bridge them into an OpenTelemetry log pipeline;
+    // if that pipeline exports through this same trace
+    // exporter, a failed export would log, produce a record, and fail again. The OpenTelemetry
+    // SDK enters this scope around its own exporters, but Datadog span export runs on
+    // libdatadog's worker instead, so we enter it ourselves. The guard is `!Send`, hence it is
+    // taken here rather than around the `async` export itself.
+    let _suppress_guard = Context::enter_telemetry_suppressed_scope();
```
hoolioh: Same as the comment above

hoolioh (approved):

PR #283, opened 2026-07-31
feat(propagation): emit and forward OTel consistent-probability ot tracestate member

MilanGarnier:
## Summary
- Parse the inbound `ot` tracestate member (rv/th) and forward it raw, rewriting only on injection
- Emit `rv`/`th` on inject based on the sampling decision, keeping unrecognized ot sub-keys
- Thread the ot decision from sampler through to the inject context

## Test plan
- [x] cargo test -p datadog-opentelemetry
- [x] cargo clippy --all-targets -- -D warnings
- Tests in https://github.com/DataDog/system-tests/pull/7372 to enable (except the ones involving force keep)

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -182,7 +182,7 @@ jobs:
   parametric:
     needs:
       - build-artifacts
-    uses: DataDog/system-tests/.github/workflows/run-parametric.yml@623d8cee11cfc63d3e5a90c1a70ce8ecf726bd3a  # main
+    uses: DataDog/system-tests/.github/workflows/run-parametric.yml@53687a2f4eab136c61d2a97474a37846d9ca6959  # main
     permissions:
       contents: read
       id-token: write
@@ -191,5 +191,5 @@ jobs:
       binaries_artifact: system_tests_binaries
       job_count: 2
       job_matrix: '[1,2]'
-      ref: 623d8cee11cfc63d3e5a90c1a70ce8ecf726bd3a # main
+      ref: 53687a2f4eab136c61d2a97474a37846d9ca6959 # main
       push_to_test_optimization: true
--- a/datadog-opentelemetry/src/propagation/context.rs
+++ b/datadog-opentelemetry/src/propagation/context.rs
@@ -117,7 +117,7 @@ pub struct InjectSpanContext<'a> {
     /// W3C tracestate data to inject.
     pub tracestate: Option<InjectTraceState>,
     /// The OTel consistent-probability `ot` member to emit, if any.
-    pub ot: Option<String>,
+    pub ot_member: Option<&'a str>,
 }
 
 #[cfg(test)]
@@ -142,7 +142,7 @@ pub(crate) fn span_context_to_inject(c: &mut SpanContext) -> InjectSpanContext<'
                 },
             ))
         }),
-        ot: c.tracestate.as_ref().and_then(|ts| ts.ot.clone()),
+        ot_member: c.tracestate.as_ref().and_then(|ts| ts.ot_member.as_deref()),
     }
 }
 
--- a/datadog-opentelemetry/src/propagation/context.rs
+++ b/datadog-opentelemetry/src/propagation/context.rs
@@ -116,6 +116,8 @@ pub struct InjectSpanContext<'a> {
     pub is_remote: bool,
     /// W3C tracestate data to inject.
     pub tracestate: Option<InjectTraceState>,
+    /// The OTel consistent-probability `ot` member to emit, if any.
+    pub ot: Option<String>,
 }
 
 #[cfg(test)]
@@ -140,6 +142,7 @@ pub(crate) fn span_context_to_inject(c: &mut SpanContext) -> InjectSpanContext<'
                 },
             ))
         }),
+        ot: c.tracestate.as_ref().and_then(|ts| ts.ot.clone()),
     }
 }
 
--- a/datadog-opentelemetry/src/propagation/tracecontext.rs
+++ b/datadog-opentelemetry/src/propagation/tracecontext.rs
@@ -62,6 +62,19 @@ pub(crate) fn ot_extract_rv(raw: &str) -> Option<u64> {
     })
 }
 
+/// Removes malformed `rv`/`th` subkeys, preserving valid and unknown subkeys.
+pub(crate) fn ot_sanitize(raw: &str) -> Option<String> {
+    let parts: Vec<_> = raw
+        .split(';')
+        .filter(|item| match item.split_once(':') {
+            Some(("rv", value)) => ot_parse_hex_56(value, Some(14)).is_some(),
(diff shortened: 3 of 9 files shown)


datadog-opentelemetry/src/text_map_propagator.rs
```diff
@@ -145,6 +148,11 @@ impl DatadogPropagator {
             &mut HashMap::new()
         };
 
+        let ot = propagation_data.ot.take().or_else(|| {
+            cx.get::<DatadogExtractData>()
+                .and_then(|extract_data| extract_data.ot.clone())
+        });
```
iunanua: It seems the fallback could resurrect an erased `ot`:

`get_trace_propagation_data` returns `EMPTY_PROPAGATION_DATA` when the trace isn't registered, so `ot: None` is indistinguishable from "the local root resolved this to None on purpose":

- inbound traceparent + deferred `dd=` +` ot=th:e6666666666666` (no rv)
- local root's decision is non-probability - e.g. the rate limiter fired, which makes `otel_consistent_sampling()` return `None` via `rl_effective_rate.is_some()` (not a corner case)
- `resolve_outbound_ot(true, None, Some("th:e666..."))` > `ot_extract_rv` > `None, ot_set_rv_th` filters the th out, no other subkeys > registry stores `ot: None`
- inject falls back to `DatadogExtractData` and emits `ot=th:e6666666666666`

MilanGarnier: That's true, 
I fixed it in [fix(sampling): difference registered and unregistered traces](https://github.com/DataDog/dd-trace-rs/pull/283/commits/e5d83c980745023026c81540014e2ddc5308d3df)
solution proposed: `get_trace_propagation_data(...) -> Option<&TracePropagationData>` instead of `&TracePropagationData` (and `TracePropagationData::default()` returns `NO_PROPAGATION_DATA`)

datadog-opentelemetry/src/propagation/context.rs
```diff
@@ -116,6 +116,8 @@ pub struct InjectSpanContext<'a> {
     pub is_remote: bool,
     /// W3C tracestate data to inject.
     pub tracestate: Option<InjectTraceState>,
+    /// The OTel consistent-probability `ot` member to emit, if any.
+    pub ot: Option<String>,
```
paullegranddc: why does this need to be a String? `origin` is already borrowed

datadog-opentelemetry/src/propagation/tracecontext.rs
```diff
@@ -55,6 +136,8 @@ pub struct Tracestate {
     pub(crate) lower_order_trace_id: Option<String>,
     pub(crate) propagation_tags: Option<HashMap<String, String>>,
     pub(crate) additional_values: Option<Vec<(String, String)>>,
+    /// Raw inbound OTel field (useful for consistent-probability member value)
```
paullegranddc: the comment doesn't add that much here

MilanGarnier: deleted it, and renamed it to `ot_member`

datadog-opentelemetry/src/propagation/context.rs
MilanGarnier: it's true that it does not need to at this stage, fixed in [refactor(propagation): rename ot to ot_member, remove comment, use](https://github.com/DataDog/dd-trace-rs/pull/283/commits/650cf88d85f56791d59310892d43b75c19bb161f)
(it still needs to be a string at injection because it may be modified)

datadog-opentelemetry/src/text_map_propagator.rs
iunanua: Thkx

iunanua:
are there any system tests for this feature?

MilanGarnier:
> are there any system tests for this feature?

@iunanua Yes! In https://github.com/DataDog/system-tests/pull/7372, I'll enable them in https://github.com/DataDog/system-tests/pull/7391

datadog-opentelemetry/src/propagation/tracecontext.rs
iunanua, line 303: I'm wondering if `ot` entry should also be filtered to avoid duplicated `ot` parts in the injected tracestate (`dd=...,ot=<resolved>,ot=<original>`)

It'd be a corner case as original tracestate should be extracted and parsed and then the `ot` part would be removed from the otel header but I guess filtering it out in `InjectTraceState::additional_values` would be safer

.github/workflows/test.yaml
```diff
@@ -180,5 +180,5 @@ jobs:
       binaries_artifact: system_tests_binaries
       job_count: 2
       job_matrix: '[1,2]'
-      ref: 271863c89eb498754525cc8a7f59f46868a729e4 # main
+      ref: 431570efc4e1991d81b2d5e7bcf20218b3e38e7a # main
```
iunanua: I guess it should be updated once https://github.com/DataDog/system-tests/pull/7391 is merged

datadog-opentelemetry/src/propagation/tracecontext.rs
```diff
@@ -37,6 +38,86 @@ const INVALID_CHAR_REPLACEMENT: char = '_';
...
+    for part in parts {
+        let sep_len = if count == 0 { 0 } else { 1 };
+        if len + sep_len + part.len() > max_len {
+            break;
+        }
+        len += sep_len + part.len();
+        count += 1;
+    }
+    (count > 0).then(|| parts[..count].join(";"))
+}
```
paullegranddc: We could remove an allocation here 

```suggestion
    let parts = rv_part
        .as_deref()
        .into_iter()
        .chain(th_part.as_deref())
        .chain(others);
    join_capped(parts, TRACESTATE_OT_KEY_MAX_LENGTH)
}

/// Joins leading `parts` with `;` capping length under `max_len`
fn join_capped<I: IntoIterator<Item = &str>(parts: I, max_len: usize) -> Option<String> {
    let mut res = String::new();
    let mut i = 0;
    for part in parts {
      let sep_len = if i == 0 { 0 } else { 1 };
      if res.len() + sep_len + part.len() > max_len {
        return Some(res);
      }
      if i != 0 { res.push(';'); }
      res.push(part);
      i += 1;
    }
    (i > 0).then_some(res) 
}
``````

paullegranddc (approved):
LGTM overall other than a small comment

.github/workflows/test.yaml
MilanGarnier: @iunanua  Some other tests cannot pass on the current version because they were hardcoding the presence of only one vendor in the tracestate, and they were updated.

datadog-opentelemetry/src/propagation/tracecontext.rs
MilanGarnier: applied, thanks !

PR #284, opened 2026-08-03
fix(otel): only error when no extractor recovers a real trace [SVLS-9550]

litianningdatadog:
# Problem

`DatadogCompositePropagator` runs every configured extractor unconditionally. Each one logs its own failure at `ERROR` independently, even when a different, later extractor recovers the trace in the same request.

A user reported that after upgrading their .NET tracer + Lambda Extension: `Propagator (tracecontext): Failed to extract traceparent: Cannot extract from traceparent, invalid traceparent` at `ERROR` on ~27% of invocations (6,814 / 24,911 requests in one day) — [SLES-2958](https://datadoghq.atlassian.net/browse/SLES-2958). Every one of those lines was immediately followed by a successful extraction via the `Datadog` headers. Tracing was never broken — this was pure log and telemetry (`add_log_error`) noise. The `Datadog` extractor's own parse failures have the same bug.

# Solution

- `Propagator::try_extract` replaces the trait's `extract` method internally. It returns `Result<SpanContext, Error>` instead of `Option<SpanContext>`, so a failure is data the composite propagator can inspect instead of a log line that already happened.
- The public `datadog::extract`/`tracecontext::extract` functions are untouched — they still log `ERROR` on failure, for any consumer calling them directly outside `DatadogCompositePropagator`. `try_extract` is a new, crate-private sibling used only by the composite path, so fixing the aggregate-logging case doesn't blind standalone callers of the public propagation API.
- `extract_available_contexts` now collects successes and failures across every configured extractor, then logs a single aggregate `ERROR` — naming which formats were tried (`tried: datadog, tracecontext`) — only when no *real* trace was recovered. A zero-trace context (B3's sampling-only `b3: 1` form) doesn't count as recovery, so it can't mask a genuine failure from a different, trace-id-bearing format that ran alongside it.

# Tests
- Run AWS-based tests using [test function](https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/functions/tianning-li-20260802-2342-SLES-2958-dotnet/versions/5?subtab=url&tab=code). 
1. [Without the fix](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Ftianning-li-20260802-2342-SLES-2958-dotnet/log-events/2026$252F08$252F03$252F$255B$2524LATEST$255D3445ef84874445e8bca0830c08883baa$3FfilterPattern$3DERROR$26start$3D1785729600000$26end$3D1785815999000), we were getting ERROR log.
2. [With the fix](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Ftianning-li-20260802-2342-SLES-2958-dotnet/log-events/2026$252F08$252F03$252F$255B$2524LATEST$255D6b73d998573a492c879cc989bff40a25$3FfilterPattern$3DERROR$26start$3D1785729600000$26end$3D1785790799000), we no longer get ERROR log if the message only has malformed traceparent
3. [With the fix](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Ftianning-li-20260802-2342-SLES-2958-dotnet/log-events/2026$252F08$252F04$252F$255B$2524LATEST$255Dde7c3577adea4287a2fbc5fcad33c773$3FfilterPattern$3DERROR$26start$3D1785801600000$26end$3D1785887999000), we do see ERROR logs if the message has malformed traceparent AND misses critical info such as  `x-datadog-trace-id`

# Additional notes

- New tests: the original malformed-`traceparent`-recovers-via-`Datadog` scenario (SLES-2958), B3's sampling-only header not masking a real Datadog failure, and direct unit tests for the extraction-outcome decision (`lost_real_trace`).
- **Observability trade-off**: the aggregate `ERROR` still fires on real trace-context loss, but the specific per-extractor reason (which header, what parse error) is now only visible in `dd_debug!` calls. dd-trace-rs's log threshold is set by a crate-private `set_max_level` that only its own `Config` init calls — a host like the Lambda Extension can't raise it. Confirmed empirically: redeploying with the extension's `DD_LOG_LEVEL=debug` produced plenty of the extension's own debug output, zero dd-trace-rs log lines. The `tried: ...` text in the aggregate message is the only mitigation left at the default level.
- Verified end-to-end on a live Lambda deployment reproducing the exact customer scenario (malformed `traceparent` + valid Datadog headers): the `ERROR` disappears, trace still resolves via the Datadog headers.

[SLES-2958]: https://datadoghq.atlassian.net/browse/SLES-2958?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SVLS-9550]: https://datadoghq.atlassian.net/browse/SVLS-9550?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/datadog-opentelemetry/src/propagation/b3.rs
+++ b/datadog-opentelemetry/src/propagation/b3.rs
@@ -396,7 +396,8 @@ mod test {
         let carrier = carrier_with("80f198ee56343ba8-00f067aa0ba902b7-1");
         let propagator = TracePropagationStyle::B3SingleHeader;
         let ctx = propagator
-            .extract(&carrier, &Config::builder().build())
+            .try_extract(&carrier, &Config::builder().build())
+            .map(Result::unwrap)
             .expect("b3 single-header dispatch should produce context");
         assert_eq!(ctx.trace_id, 0x80f1_98ee_5634_3ba8);
     }
--- a/datadog-opentelemetry/src/propagation/b3multi.rs
+++ b/datadog-opentelemetry/src/propagation/b3multi.rs
@@ -375,7 +375,8 @@ mod test {
         ]);
         let propagator = TracePropagationStyle::B3Multi;
         let ctx = propagator
-            .extract(&carrier, &Config::builder().build())
+            .try_extract(&carrier, &Config::builder().build())
+            .map(Result::unwrap)
             .expect("b3multi dispatch should produce context");
         assert_eq!(ctx.trace_id, 0x80f1_98ee_5634_3ba8);
     }
--- a/datadog-opentelemetry/src/propagation/datadog.rs
+++ b/datadog-opentelemetry/src/propagation/datadog.rs
@@ -191,22 +191,39 @@ fn validate_tag_value(value: &str) -> bool {
 }
 
 /// Extract trace context from a carrier using Datadog headers.
+///
+/// `DatadogCompositePropagator` calls `try_extract` instead; this is for direct
+/// callers of the public propagation API (`_unstable_propagation`/`test-utils`).
+#[allow(dead_code)]
 pub fn extract(
     carrier: &dyn Extractor,
     config: &(impl PropagationConfig + ?Sized),
 ) -> Option<SpanContext> {
-    let lower_trace_id = match extract_trace_id(carrier) {
-        Ok(trace_id) => trace_id?,
+    match try_extract(carrier, config)? {
+        Ok(context) => Some(context),
         Err(e) => {
             dd_error!("Propagator (datadog): Error extracting trace_id {e}");
-            return None;
+            None
         }
+    }
+}
+
+/// Same as [`extract`], but returns the trace-id error instead of logging it. The
+/// composite propagator uses this so it can log once, after checking if another
+/// format recovered.
+pub(crate) fn try_extract(
+    carrier: &dyn Extractor,
+    config: &(impl PropagationConfig + ?Sized),
+) -> Option<Result<SpanContext, Error>> {
+    let lower_trace_id = match extract_trace_id(carrier) {
+        Ok(trace_id) => trace_id?,
+        Err(e) => return Some(Err(e)),
     };
 
     let parent_id = match extract_parent_id(carrier) {
         Ok(parent_id) => parent_id.unwrap_or_default(),
         Err(e) => {
-            dd_error!("Propagator (datadog): Error extracting parent_id {e}");
+            dd_warn!("Propagator (datadog): Error extracting parent_id {e}");
             0
         }
     };
@@ -239,7 +256,7 @@ pub fn extract(
         tags.get(DATADOG_HIGHER_ORDER_TRACE_ID_BITS_KEY),
     );
 
-    Some(SpanContext {
+    Some(Ok(SpanContext {
         trace_id,
         span_id: parent_id,
         sampling,
@@ -248,7 +265,7 @@ pub fn extract(
         links: Vec::new(),
         is_remote: true,
         tracestate: None,
-    })
+    }))
 }
 
 fn extract_trace_id(carrier: &dyn Extractor) -> Result<Option<u64>, Error> {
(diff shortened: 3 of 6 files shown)


iunanua:
Hi, my biggest concern is that if an app uses two propagation styles and sends a malformed header for one of them, it wouldn't be noticed until debug logs were enabled.

Idk if adding a telemetry metric would be enough. There is a `context_header_style.malformed` metric tagged by header style.

And to address the 'sampling-only B3' problem stated by codex, we could change the Propagator trait so that method `extract` returns something like 
```rust
  struct ExtractionOutcome {
      context: Option<SpanContext>,
      error: Option<ExtractionError>,
  }
```
Then we could just check for errors.

litianningdatadog:
> Hi, my biggest concern is that if an app uses two propagation styles and sends a malformed header for one of them, it wouldn't be noticed until debug logs were enabled.
> 
> Idk if adding a telemetry metric would be enough. There is a `context_header_style.malformed` metric tagged by header style.
> 
> And to address the 'sampling-only B3' problem stated by codex, we could change the Propagator trait so that method `extract` returns something like
> 
> ```rust
>   struct ExtractionOutcome {
>       context: Option<SpanContext>,
>       error: Option<ExtractionError>,
>   }
> ```
> 
> Then we could just check for errors.

@iunanua Agreed. This a real gap — once one style recovers a real trace, a malformed header on another only logs at debug. Deliberate tradeoff here, but it can hide a broken integration until the other style fails too.

`context_header_style.malformed` doesn't exist yet. Our telemetry client only has a fixed macro-generated metric enum, no dynamic per-style tags — adding it is more than this PR. I would defer this for a followup PR. 

`ExtractionOutcome` would be the better shape — `Result` can't hold a context and an error together, which is exactly this case. But it's a trait signature change across every extractor, same as above, I would prefer a followup PR to keep this PR stay focused.

iunanua:
> `context_header_style.malformed` doesn't exist yet. Our telemetry client only has a fixed macro-generated metric enum, no dynamic per-style tags — adding it is more than this PR. I would defer this for a followup PR.

It's just a matter of defining the metric (in telemetry.rs where the others are defined) for each style we're interested in:
```rust
ContextHeaderStyleMalformedTracecontext => ("context_header_style.malformed", Tracers, Count, ["header_style" => "tracecontext"]),
ContextHeaderStyleMalformedDatadog => ("context_header_style.malformed", Tracers, Count, ["header_style" => "datadog"]),
    
```
and then call `telemetry::add_point` where you handle the ~~`lost_real_trace`~~  the error returned by `try_extract`

litianningdatadog:
> > `context_header_style.malformed` doesn't exist yet. Our telemetry client only has a fixed macro-generated metric enum, no dynamic per-style tags — adding it is more than this PR. I would defer this for a followup PR.
> 
> It's just a matter of defining the metric (in telemetry.rs where the others are defined) for each style we're interested in:
> 
> ```rust
> ContextHeaderStyleMalformedTracecontext => ("context_header_style.malformed", Tracers, Count, ["header_style" => "tracecontext"]),
> ContextHeaderStyleMalformedDatadog => ("context_header_style.malformed", Tracers, Count, ["header_style" => "datadog"]),
>     
> ```
> 
> and then call `telemetry::add_point` where you handle the `lost_real_trace`

@iunanua That's fair. The enum variants are cheap. However, `propagation/mod.rs` has no dependency on `core::telemetry` as of today. Wiring this in adds that edge, plus a real decision on emission granularity (per-failure in the `failures` vec vs. only when `lost_real_trace` fires), plus tests for it. That's small but it's still new scope beyond fixing the log-flooding false positive. Hence a separate PR would be better for this move.

bantonsson (approved):

PR #294, opened 2026-08-21
chore(ci): add cargo deny advisories check and unmaintained advisory config

bantonsson:
# What does this PR do?

Add a non-blocking cargo deny job to lint.yaml that runs advisories, bans, and sources checks on both the root and instrumentation workspaces. 

# Motivation

We should be aware of the state of the crates we depend on. This job complements the blocking cargo-deny-bans gate with visibility into vulnerabilities and unmaintained crates without failing the workflow.

# Additional Notes

Look for the comment.

### DIFF
--- a/.github/workflows/cargo-deny.yaml
+++ b/.github/workflows/cargo-deny.yaml
@@ -1,0 +1,114 @@
+name: cargo-deny/advisories
+# Non-blocking advisory check that posts a PR comment with the results.
+# Kept separate from lint.yaml (push-only) so adding PR commenting does not
+# duplicate every lint job for each PR update, and separate from the blocking
+# cargo-deny/bans gate which always blocks on banned crates.
+on:
+  pull_request:
+permissions:
+  contents: read
+  pull-requests: write
+# Cancel stale runs so an older run cannot overwrite a newer comment
+concurrency:
+  group: cargo-deny-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+env:
+  CARGO_TERM_COLOR: always
+  # Set to true to make cargo deny errors fail the job
+  FAIL_IF_CARGO_DENY: false
+jobs:
+  cargo-deny-advisories:
+    runs-on: ubuntu-latest
+    name: "cargo-deny/advisories"
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: ./.github/actions/cache
+        with:
+          rust_version: stable
+      - name: Install stable toolchain
+        run: rustup install stable && rustup default stable
+      - name: Install cargo-deny
+        uses: taiki-e/install-action@a2a5f6e99e1a31540baa0468acfa302cff0f359f # v2.86.4
+        with:
+          tool: cargo-deny
+      - name: Run cargo deny
+        id: cargo-deny
+        run: |
+          set +e
+          ROOT_OUTPUT=$(cargo deny --color never --log-level error check advisories bans sources 2>&1)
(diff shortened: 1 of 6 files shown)


hoolioh (approved):

.github/workflows/cargo-deny.yaml
```diff
@@ -0,0 +1,114 @@
...
+        run: rustup install stable && rustup default stable
+      - name: Install cargo-deny
+        uses: taiki-e/install-action@a2a5f6e99e1a31540baa0468acfa302cff0f359f # v2.86.4
+        with:
+          tool: cargo-deny
+      - name: Run cargo deny
+        id: cargo-deny
+        run: |
+          set +e
+          ROOT_OUTPUT=$(cargo deny --color never --log-level error check advisories bans sources 2>&1)
```
ekump, line 39: I notice we do the same thing in libdatadog, we set the log level to error but below we count and report warnings too. Any reason we don't set the log-level to warning?

deny.toml
```diff
@@ -1,3 +1,5 @@
+[advisories]
+
```
ekump, line 2: I think you want this so you only get warned on direct dependencies?
```suggestion
[advisories]
unmaintained = "workspace"  
```

.github/workflows/cargo-deny.yaml
```diff
@@ -0,0 +1,114 @@
...
+      - name: Install stable toolchain
+        run: rustup install stable && rustup default stable
+      - name: Install cargo-deny
+        uses: taiki-e/install-action@a2a5f6e99e1a31540baa0468acfa302cff0f359f # v2.86.4
+        with:
+          tool: cargo-deny
+      - name: Run cargo deny
+        id: cargo-deny
+        run: |
+          set +e
```
ekump, line 38: is this too broad? If anything fails at all it'll still report success.

.github/workflows/cargo-deny.yaml
```diff
@@ -0,0 +1,114 @@
...
+            JOB_URL="${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+          fi
+
+          BODY="${BODY}\n---\n*Updated: $(date -u '+%Y-%m-%d %H:%M:%S UTC') | Commit: ${{ github.sha }} | [cargo deny job results](${JOB_URL})*"
+
+          echo -e "$BODY" > cargo-deny-results.md
+      - name: Find existing comment
+        uses: peter-evans/find-comment@3eae4d37986fb5a8592848f6a574fdf654e61f9e # v3.1.0
+        id: find-comment
+        if: ${{ !github.event.pull_request.head.repo.fork }}
```
ekump, line 97: nit: I think this will still run cargo deny, it just won't post the result. Could you just skip the whole run on fork?

.github/workflows/cargo-deny.yaml
bantonsson: I have no idea. I'm just copying the template.

deny.toml
bantonsson:
```
error[deprecated]: this key has been removed, see https://github.com/EmbarkStudios/cargo-deny/pull/611 for migration information
  ┌─ /Users/bjorn.antonsson/go/src/github.com/DataDog/dd-trace-rs/deny.toml:2:1
  │
2 │ unmaintained = "workspace"
  │ ━━━━━━━━━━━━
```

.github/workflows/cargo-deny.yaml
bantonsson: I'll take a look at it.

deny.toml
bantonsson: Too many old versions of cargo deny... I'll fix this.

bantonsson:
Fixes here https://github.com/DataDog/dd-trace-rs/pull/295

PR #297, opened 2026-08-26
ci(licenses): validate license CSVs across workspaces

ygree:
## What does this PR do?

Updates third-party license CSV generation and validation to cover both the root Cargo workspace and the `instrumentation` workspace.

- Adds `instrumentation/LICENSE-3rdparty.csv`
- Updates the license CI job to generate and diff both CSV files from temporary outputs
- Keeps CI close to the previous flow by calling `dd-rust-license-tool` directly
- Updates `scripts/update_license_3rdparty.sh` to regenerate both workspace CSVs together
- Generates local script outputs in a temporary directory before copying them into tracked files
- Updates the license Docker image to generate both CSVs at build time with cache mounts
- Documents the simplified `./scripts/update_license_3rdparty.sh` workflow

## Motivation

The repository now has multiple Cargo workspaces with separate dependency graphs. CI previously validated only the root `LICENSE-3rdparty.csv`, so dependencies under `instrumentation` were not covered by the license CSV check.

## Additional Notes

No Rust runtime code is changed. The CI failure artifact now includes both generated license CSV files to make remediation easier.

### DIFF
--- a/scripts/Dockerfile.license
+++ b/scripts/Dockerfile.license
@@ -14,5 +14,10 @@ WORKDIR /build
 
 COPY . .
 
-ENTRYPOINT ["dd-rust-license-tool"]
-CMD ["--manifest-path", "Cargo.toml", "dump"]
+RUN --mount=type=cache,target=/usr/local/cargo/registry \
+    --mount=type=cache,target=/usr/local/cargo/git \
+    mkdir -p /licenses/instrumentation && \
+    dd-rust-license-tool dump > /licenses/LICENSE-3rdparty.csv && \
+    dd-rust-license-tool --manifest-path instrumentation/Cargo.toml dump > /licenses/instrumentation/LICENSE-3rdparty.csv
+
+ENTRYPOINT ["/bin/sh", "-ec", "test \"$#\" -eq 1; exec cat \"$1\"", "--"]
--- a/scripts/Dockerfile.license
+++ b/scripts/Dockerfile.license
@@ -14,8 +14,5 @@ WORKDIR /build
 
 COPY . .
 
-RUN --mount=type=cache,target=/usr/local/cargo/registry \
-    --mount=type=cache,target=/usr/local/cargo/git \
-    dd-rust-license-tool dump > /licenses.csv
-
-ENTRYPOINT ["cat", "/licenses.csv"]
+ENTRYPOINT ["dd-rust-license-tool"]
+CMD ["--manifest-path", "Cargo.toml", "dump"]
--- a/scripts/update_license_3rdparty.sh
+++ b/scripts/update_license_3rdparty.sh
@@ -12,17 +12,9 @@ ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && cd .. && pwd)"
 
 usage() {
     cat <<EOF
-Usage: $0 WORKSPACE_PATH [WORKSPACE_PATH...]
+Usage: $0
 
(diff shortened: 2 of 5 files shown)


scripts/Dockerfile.license
```diff
@@ -14,8 +14,5 @@ WORKDIR /build
 
 COPY . .
 
-RUN --mount=type=cache,target=/usr/local/cargo/registry \
-    --mount=type=cache,target=/usr/local/cargo/git \
-    dd-rust-license-tool dump > /licenses.csv
-
-ENTRYPOINT ["cat", "/licenses.csv"]
+ENTRYPOINT ["dd-rust-license-tool"]
```
ekump: why isn't this using the cache for /usr/local/cargo anymore?

scripts/update_license_3rdparty.sh
```diff
@@ -10,13 +10,79 @@ TOOL_VERSION="1.0.6"
...
+    output="${workspace_path}/LICENSE-3rdparty.csv"
+
+    if [ ! -f "${manifest}" ]; then
+        echo "ERROR: ${workspace_path} is not a Cargo workspace directory."
+        echo "Expected manifest at: ${manifest}"
+        exit 1
+    fi
+
+    echo "Generating ${output}..."
+    dd-rust-license-tool --manifest-path "${manifest}" dump > "${output}"
```
ekump: If this fails part of the way through it's going to dump something incomplete. Shouldn't we only overwrite the file once it successfully generates? (also applies [here](https://github.com/DataDog/libdatadog-nodejs/pull/188)).

scripts/update_license_3rdparty.sh
```diff
@@ -10,13 +10,79 @@ TOOL_VERSION="1.0.6"
...
+
+Each WORKSPACE_PATH must point to a Cargo workspace directory. The generated file is written next
+to that workspace's Cargo.toml as LICENSE-3rdparty.csv.
+
+Examples:
+  $0 .
+  $0 instrumentation
+  $0 . instrumentation
+EOF
+}
```
ekump: nit: Probably worth pointing out that the path must be relative, not absolute.

scripts/update_license_3rdparty.sh
ekump, line 1: There is no separate [licesnse-tool.toml](https://github.com/DataDog/dd-trace-rs/blob/12c952b0c26905de6b77c10a16e8fac901e885ed/license-tool.toml) for `instrumentation/`. I think that's a good thing. But, that means you can't run the script form within `instrumentation/`. It's probably worth documenting that somewhere.

ekump (approved):
minor comments, but overall LGTM

scripts/Dockerfile.license
ygree: Good catch! I've overseen its importance when dd-rust-license-tool is not installed.

scripts/update_license_3rdparty.sh
ygree: I decided to make the script to update both license files to keep it simple.

scripts/update_license_3rdparty.sh
ygree: I decided to make the script to update both license files to keep it simple.

scripts/update_license_3rdparty.sh
ygree: Good point! I'll leave it as it was before rather than updating it in place.

ekump (approved):

PR #304, opened 2026-09-21
fix: use trace buffer wait method instead of adding a wrapper

paullegranddc:
# Motivation

We currently have quite complex code to wait for span flusingh to be completed in the trace buffer.

This is not necessary, as a flush_and_wait method was added in libdatadog.

This PR removes the abstraction around the trace buffer and use the native method instead.

# Changes

* Remove PendingSpans code
* Call flush_and_wait instead


iunanua (approved):

yannham (approved):

bantonsson (approved):
I got such a weird deja vu from this PR, and lo and behold the code was still in a local branch... 🤦🏼‍♂️

datadog-opentelemetry/src/span_exporter.rs
```diff
@@ -512,7 +416,6 @@ impl Export<BufferedSpan> for SpanDataExport {
     > {
         // Account for every span we drained from the buffer, regardless of whether the export
         // ultimately succeeds — `send_chunk` already counted them on entry.
```
bantonsson: Isn't this comment wrong now?

```suggestion
```

paullegranddc: Yes, I have changes stacked on top of this and missed to commit this deletion

